Issue #6, opened 2012-04-23
observe code segfaults if sweeps that save do not end at corner


g1257:
This should not be a problem because all previous data must have already being processed and printed by observe.
Some have noticed, however, that because the printing is probably buffered, sometimes the segfault precludes the printing from happening at all, particularly when running observe in a batch job.
WORKAROUND: Have the sweeps that save(_) to disk end at one of the corners.
(_) Sweeps that save are those that have the first bit of the 3rd number of the finite loop triplet set.

PR #44, opened 2025-10-17
Cmake for dmrgpp

PDoakORNL:
I think this does what the perl build system does and uses only DmrgDriver.pm plus a couple of little stub scripts at configure and build time.  It depends on PsimagLite PR #6 and the build fix here at #43 .

I think this will require some more iterations but it built on what i did 5-6 years ago and is a little more flexible wrt PsimagLite.

I used a cmake line like this 

``` shell
% cmake -DPsimagLite_DIR=/raid/epd/codes/pml_clean/build/ -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DBLA_VENDOR=OpenBLAS -GNinja ..
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,15 +28,12 @@ execute_process(
   ERROR_VARIABLE perl_error
   RESULT_VARIABLE perl_result
   OUTPUT_STRIP_TRAILING_WHITESPACE)
-message("perl error ${perl_error}")
 string(REGEX MATCH "([0-9]+)" match_result ${perl_error})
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -20,7 +20,6 @@ BraceWrapping:
   AfterObjCDeclaration: true
   AfterStruct:     true
   AfterUnion:      true
-  AfterExternBlock: false
```
dalg24: Please no Clang-Format changes in this PR.  Let us not creep the scope and only tackle setting up initial cmake support.

PDoakORNL:
PR for the wrong branch.

PR #50, opened 2025-10-31
Link modifiers now have site dependence

g1257:
Changes to code logic; Link modifiers now have two modes.

Model writers can pass a modifier to a Hamiltonian term (connection). This modifier is a C++ lambda with signature
```
(const ComplexOrRealType& connector_value)
```
or
```
(const ComplexOrRealType& connector_value, RealType time, unsigned int site)
```
and in both cases returns the modified connector value.

This PR adds the second option, and drops the old option where connector_value and time were allowed.

### DIFF
--- a/src/Engine/HamiltonianAbstract.h
+++ b/src/Engine/HamiltonianAbstract.h
@@ -1,6 +1,7 @@
 #ifndef HAMILTONIANABSTRACT_H
 #define HAMILTONIANABSTRACT_H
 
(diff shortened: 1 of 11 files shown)


g1257:
@dalg24  This is needed for the work that Steve Johnston is doing.
I wonder if you don't mind reviewing this, at least making sure it compiles on your end.

dalg24:
I was able to build with the changes from #51

PR #51, opened 2025-11-07
CMake: Drop unwarranted compile options

dalg24:
Do not mess with options to request warnings.
Do not add compile options to set the build type either.


g1257:
Can you run?
Like
cd dmrgpp/build
src/dmrg -f ../TestSuite/inputs/input2.ain

because I need -DUSE_BOOST for DMRG++

g1257:
If we are ready I'll merge #50  and then #51 
otherwise, please let me know.

PR #55, opened 2025-11-13
Embed PsimagLite build into dmrgcpp

dalg24:
Use `add_subdirectory(PsimagLite)` instead of searching for an external dependency.
Define a `psimaglite::psimaglite` alias target so we can refer to it just as if it was imported.


PDoakORNL:
This is fine but I fails on the first build with ninja because GitRevision.h is missing. I prefer mine in #56 but as long as I am not responsible for cleaning up PsimagLite at a later date I don't care much.

PR #58, opened 2025-11-14
Add Catch2

PDoakORNL:
Adding [Catch2](https://github.com/catchorg/Catch2)
Basic integration in dmrgpp cmake.
Added example unit test
/PsimagLit/src/tests/test_AkimaSpline.cpp


dalg24 (commented):
That's a lot of code to embed.
At the very least I would want us to drop everything that is not absolutely necessary from the Catch2 subdirectory.

But I would prefer if we followed the FetchContent suggestion in
https://github.com/catchorg/Catch2/blob/a1faad9315ece8e7146e9d2263ceb3d42ea0619a/docs/cmake-integration.md#cmake-targets

dalg24:
FetchContent supports attempting a regular find_package search and then falling back to downloading.

PR #60, opened 2025-11-14
Improve build instructions

dalg24:


PDoakORNL (approved):
Not the way I do it but valid.

I tend to do this
```
mkdir build; cd build
cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpic++ -GNinja ..
ninja
ctest --output-on-failure
```
This is my habit since I'm often in environments with gcc, clang, and mpi wrappers are available in the path at one time. It's not uncommon for clusters and hpc machines to be similar.

since I use different compilers different days and stick to openmpi built with the compiler I'm using this seems like good habit to avoid confusing issues.
cmake does its best but I get scary stuff like system gcc, clang openmpi, etc. various packages and modules seem to try and set CC and CXX to whatever they like which cmake is sensitive to as well.

But I'm not sure what to tell users in general to do.

PR #61, opened 2025-11-14
Fetch Catch2

PDoakORNL:
Adding [Catch2](https://github.com/catchorg/Catch2)
Using modern cmake FetchContent with optional find package when cmake fetching is not possible.
Basic integration in dmrgpp cmake.

* Added example unit test -- /PsimagLit/src/tests/test_AkimaSpline.cpp
* For how to setup of a test program target. -- /PsimagLit/src/tests/CMakeLists.txt

Personally I think more test executables with fewer dependencies are better. This will of course be hard if you application is structured as one mega library target and one mega application target. Why is it structured that way?

This makes Incremental Builds way more efficient, if you need a reproducer for a vendor complexity is already low. Ideally you can write a unit test, fix a bug and only be compiling a few files at a time. This can take your write compile test debug  loop to <1 minute instead of ~10 or ~100 minutes.

Catch2 tests will build much more quickly if you use only the headers you need and not the single header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,7 @@ FetchContent_Declare(
   Catch2
   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
   GIT_TAG v3.11.0
-  SYSTEM
-  FIND_PACKAGE_ARGS NAMES Catch2)
+  SYSTEM FIND_PACKAGE_ARGS NAMES Catch2)
 FetchContent_MakeAvailable(Catch2)
 include(Catch)
 


PsimagLite/.clang-format
dalg24: Please withdraw that change.  It does not belong to this PR.

PDoakORNL: Yes this slipped in. Will remove

PsimagLite/src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_executable(test_AkimaSpline test_AkimaSpline.cpp)
+target_link_libraries(test_AkimaSpline PRIVATE Catch2::Catch2WithMain psimaglite)
```
dalg24: We want to always refer to the qualified library so that in principle we could use an external PsimagLite
```suggestion
target_link_libraries(test_AkimaSpline PRIVATE Catch2::Catch2WithMain psimaglite::psimaglite)
```

dalg24 (commented):
Just a few minor comments.
Looks good other than that.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.28)
+cmake_minimum_required(VERSION 3.31)
```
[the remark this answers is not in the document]
PDoakORNL: I didn't want to deal with the possible changes in the fetch content behavior between <3.30 and >3.30 
see [CMP0168](https://cmake.org/cmake/help/latest/policy/CMP0168.html#policy:CMP0168)

CMakeLists.txt
```diff
@@ -15,6 +15,15 @@ include(HandleHDF5)
 find_package(BLAS)
 find_package(LAPACK)
 
+include(FetchContent)
+FetchContent_Declare(
+  Catch2
+  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+  GIT_TAG v3.11.0
```
dalg24:
```suggestion
  GIT_TAG v3.11.0
  SYSTEM
```

PsimagLite/src/tests/test_AkimaSpline.cpp
```diff
@@ -0,0 +1,21 @@
...
+	std::vector<double> x { 1.0, 2.0, 3.0, 4.0, 5.0 };
+	std::vector<double> y { 1.0, 4.0, 9.0, 16.0, 25.0 };
+	AkimaSpline akima_spline(x, y);
+	CHECK(akima_spline(1.0) == Approx(1.0));
+	CHECK(akima_spline(2.0) == Approx(4.0));
+	CHECK(akima_spline(3.0) == Approx(9.0));
+	CHECK(akima_spline(3.5) == Approx(12.25));
+	CHECK(akima_spline(4.0) == Approx(16.0));
+	CHECK(akima_spline(5.0) == Approx(25.0));
+	CHECK_THROWS(akima_spline(25.0));
```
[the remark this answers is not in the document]
dalg24: Ping

PR #63, opened 2025-11-26
Create library interface for cincuenta to use

g1257:
We need to support cincuenta at https://github.com/g1257/cincuenta
This is a first attempt.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,9 +74,9 @@ endif()
 
 # DMRG++ as a library so cincuenta can use it
 # We may instead bring cincuenta into this repo
-add_library(dmrgpp STATIC $<TARGET_OBJECTS:Common> $<TARGET_OBJECTS:Su2Related>)
-add_library(extralinkage STATIC Qn.cpp ProgramGlobals.cpp)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -72,6 +72,12 @@ if(MPI_FOUND)
   target_link_libraries(dmrg PUBLIC MPI::MPI_CXX)
 endif()
 
+# DMRG++ as a library so cincuenta can use it
+# We may instead bring cincuenta into this repo
+add_library(dmrgpp STATIC $<TARGET_OBJECTS:Common> $<TARGET_OBJECTS:Su2Related>)
+add_library(extralinkage STATIC Qn.cpp ProgramGlobals.cpp)
+target_include_directories(extralinkage PRIVATE ${CMAKE_CURRENT_BINARY_DIR} Engine ../PsimagLite ../PsimagLite/src)
```
PDoakORNL: If extralinkage is dependent on headers from PsimagLite then I think it should be getting them through a 
```cmake
target_link_libraries(extralinkage PRIVATE psimaglite)
```
unless it generates headers then it shouldn't need `${CMAKE_CURRENT_BINARY_DIR}`  and `Engine` is a subdirectory of dmrg so I isn't likely to need that either.  Hopefully you can just drop the `target_include_directories`

PDoakORNL (approved):
I think this fine as is

PR #65, opened 2025-11-26
Adds Arnoldi Shift-And-Invert Method to PsimagLite

g1257:
This is ArnoldiIteration and Arnoldi Shift-And-Invert or SaI. The SaI uses ArnoldiIteration by containing an object of class ArnoldiIteration.

There's two tests one for each. Note that coverage isn't full by any means.

ArnoldiSaI implements a virtual inheritance based on MatrixSolverBase.hh which is also added.
In another PR, I'll move LanczosSolver and DavidsonSolver both to implement MatrixSolverBase.hh as well.


PsimagLite/examples/CMakeLists.txt
```diff
@@ -4,3 +4,9 @@ add_executable(testLanczos testLanczos.cpp)
 target_link_libraries(testLanczos psimaglite)
 
 add_test(NAME testLanczos COMMAND testLanczos)
+
+
+# testArnoldiSaI executable
+add_compile_options(-Wall -Werror -g)
```
PDoakORNL: CMake will add `-g` as appropriate for the build type.  But if you want to just apply this to all builds for this  test use
`target_compile_options(testArnoldSai PRIVATE -Wall -Werror -g)`

PsimagLite/src/Matrix.h
```diff
@@ -54,6 +54,7 @@ class Matrix {
 public:
 
 	typedef T value_type; // legacy name
+	using VectorType = std::vector<T>;
```
dalg24: This looks like a drive-by change that is not actually used anywhere.
Please revert that change.

PsimagLite/src/ArnoldiIteration.hh
```diff
@@ -0,0 +1,183 @@
...
+					error_ret = k; // k is assured to be greater than zero
+				}
+
+				double factor = (is_zero) ? 1. : 1. / denom;
+				q[k + 1] <= factor* v;
+			}
+		}
+
+		std::cout << "ArnolidIteration: done after " << q.size() << " steps;";
+		std::cout << " error=" << denom << "\n";
```
dalg24: Did you mean to commit this?
Should the printing to the standard output only happen in verbose/debug mode?

PsimagLite/src/Matrix.h
g1257: Reverted. Yes, it was drive-by-change :smile:

PsimagLite/src/ArnoldiIteration.hh
```diff
@@ -0,0 +1,183 @@
...
+	{
+		SizeType n_arnoldi = h.rows();
+		if (n_arnoldi != h.cols()) {
+			throw RuntimeError("diagHessenberg(): Hessenberg not square\n");
+		}
+
+		eigs_of_h.resize(n_arnoldi);
+		h_right_eigenvectors.resize(n_arnoldi, n_arnoldi);
+		PsimagLite::Matrix<ComplexOrRealType> vl_unused(n_arnoldi, n_arnoldi);
+		// PsimagLite::Matrix<ComplexOrRealType> h_left_eigenvectors(n_arnoldi, n_arnoldi);
```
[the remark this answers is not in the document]
g1257: Deleted.

PsimagLite/examples/testArnoldiSaI.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+// void makeItReal(std::vector<std::complex<double>>& ev)
+// {
+// 	unsigned int n = ev.size();
+// 	if (n == 0) return;
+// 	std::complex<double> factor = findFactor(ev[0]);
+// 	for (SizeType i = 0; i < n; ++i) {
+// 		ev[i] *= factor;
+// 	}
+// }
```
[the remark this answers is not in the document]
g1257: Deleted.

PsimagLite/examples/testArnoldiSaI.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+void test_short()
+{
+}
```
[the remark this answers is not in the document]
g1257: I remove this empty function. But there's going to be a part II for Arnoldi shift-and-invert, which will do the shift-and-invert part; this PR does only Arnoldi iteration proper.

PsimagLite/examples/testArnoldiSaI.cpp
```diff
@@ -0,0 +1,187 @@
...
+	PsimagLite::Matrix<ComplexOrRealType> h_right_eigenvectors;
+	arnoldi.diagHessenberg(eigs_of_h, h_right_eigenvectors, h);
+	std::cout << "Eigs of h\n";
+	std::cout << eigs_of_h << "\n\n";
+
+	std::vector<ComplexOrRealType> l_eigenvector_estimate(n);
+	arnoldi.getEigenvector(l_eigenvector_estimate, h_right_eigenvectors, q, 0);
+	// makeItReal(l_eigenvector_estimate);
+	std::cout << "\nEstimated eigenvector \n";
+	std::cout << l_eigenvector_estimate << "\n";
```
dalg24: It would be preferable to go ahead and use Catch2 in this test and actually assert what these eigenvalues and eigenvectors are expected to be instead of just printing to the screen.
Any good reason to defer that work?

PsimagLite/examples/testArnoldiSaI.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+//---------------------------------------------------------------------------//
+/*!
+ * \brief Lowest element that is also real and positive
+ *
+ * \param[in] eigs A vector of complex numbers
+ *
+ * \returns the lowest element that is also real and positive
+ */
+unsigned int findTheOne(const std::vector<std::complex<double>>& eigs)
```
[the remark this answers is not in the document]
g1257: OK. I will rework this thing then. The idea of this PR is to get some code in because there's more coming next on this same very topic, and I don't want to have a very large PR.

PsimagLite/src/SparseConversionCrs.hh
```diff
@@ -0,0 +1,22 @@
...
+#include "CrsMatrix.h"
+#include "Matrix.h"
+
+namespace PsimagLite {
+
+template <typename T> class SparseConversions {
+public:
+
+	using value_type = typename T::value_type;
+
```
dalg24: Not used?
```suggestion
```

PsimagLite/src/SparseConversionCrs.hh
```diff
@@ -0,0 +1,22 @@
+#ifndef SPARSECONVERSIONCRS_HH
+#define SPARSECONVERSIONCRS_HH
+#include "CrsMatrix.h"
+#include "Matrix.h"
```
dalg24: I think technically we only need the to include the CrsMatrix header (not blocking)

dalg24 (commented):
Please fix your assertions they should read
```C++
CHECK_THAT(value, matcher);

PR #66, opened 2025-12-05
Full pipeline tests

g1257:
This MR adds full pipeline tests for the dmrg executable. Other MRs in this series will add tests for the observe, the toolboxdmrg and manyomegas executables.

All tests test the lowest energy of the last iteration and one measurement.

Some tests may test additional quantities.

TODO for this MR:

(1) cmake-format the src/CMAkeLists.txt file because it fails the CI right now

(2) Convert missing old format  .inp inputs to  Ainur .ain inputs, and add needed ones to tests

(3) Create a cmake variable called DO_HEAVY_TESTS to allow tests that take more than ~10 seconds each but no more than 40 seconds.

(4) Separate code that runs for each input after it completes: Read from the output of each test the last lowest energy and compare it with some precision: if it is the same then return 0, else return a number other than zero.

(5) Additional measurements for some of the inputs.

### DIFF
--- a/PsimagLite/scripts/checkValueOfLabel.pl
+++ b/PsimagLite/scripts/checkValueOfLabel.pl
@@ -8,8 +8,9 @@ my ($dir) = @_;
 defined($dir) or $dir = ".";
 my ($EC_UNDEFINED, $EC_NUMERIC, $EC_STRING) = (-1, -2, -3);
 my $test2 = {n => 2, 'Found lowest eigenvalue' => -19.6759};
+my $test3 = {n => 3, 'Found lowest eigenvalue' => -21.5102};
 
-my @tests = ($test2);
+my @tests = ($test2, $test3);
 
 my $flag = 0;
 for my $test (@tests) {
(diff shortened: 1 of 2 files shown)


dalg24 (commented):
I think it look good.
You just need to fix the format of the CMakeLists.txt file (see the pre-commit CI build failure)

Just for information, how slow are we talking for input8?  What about input1818 and 1819?  We can introduce later an configuration option to enable them.

g1257:
Any ideas how to fix the two pre-commit errors. I can't figure out if it's producing a diff I should use, or if there is a cmake-format command I should use?

g1257:
Please note test energy2to3
https://github.com/g1257/dmrgpp/blob/full_pipeline_tests/src/CMakeLists.txt#L133
I've written a script that goes through tests 2 ... 3 and looks at the .cout files produced by these tests, and compares against gold or reference results.

TODO: If you want to take it up. ACTION ITEM. In another PR, one can add all other tests:
Edit https://github.com/g1257/dmrgpp/blob/full_pipeline_tests/PsimagLite/scripts/checkValueOfLabel.pl
`
my $test3 = {n => 3, 'Found lowest eigenvalue' => -21.5102};
my $test4 = {n => 4, 'Found lowest eigenvalue' => add number here}; // <--- your addition

my @tests = ($test2, $test3, $test4); // <-- add the test
`

And that should be done for all current tests. 
The gold reference or correct number can be found if you run the test and get an error, it'll say gold should be this value, that you can then copy into the script.

If you don't like the script, something else can be written instead.
Feedback very welcomed.

PR #68, opened 2025-12-09
Cleanup cmake remove unnecessary HDF5 and Loki handling in DMRG++

dalg24:
The are found in PsimagLite and come transitively when linking against psimaglite::psimaglite


g1257:
This is good; I just checked it.

PR #70, opened 2025-12-10
Resolve unused but set warning about error code returned by MPI_Bcast

dalg24:
We are getting warning about not using the error code value returned by `MPI_Bcast`.
```
/path/to/dmrgpp/PsimagLite/src/MpiYes.h:386:7: warning: variable 'errorCode' set but not used [-Wunused-but-set-variable]
  386 |                 int errorCode = MPI_Bcast(&t1, 1, datatype, root, mpiComm);
      |                     ^
```
Here I opted for suppressing the warning via a C-style cast to void to tell the compiler the value is unused but we could also call `checkError` if you prefer.


g1257:
Yes, this is good. Merge anytime.

g1257 (approved):

PR #73, opened 2025-12-12
Adds Psimaglite tests

g1257:
This PR adds PsimagLite tests in PsimagLite/examples

Note that we need to convert the results which are now printed into comparisons with reference results, and this needs to be done with Catch2, so this tests will actually become unit tests. This work should not be done in this MR but in a subsequent one.

### DIFF
--- a/PsimagLite/examples/CMakeLists.txt
+++ b/PsimagLite/examples/CMakeLists.txt
@@ -1,196 +1,190 @@
-# CMakeLists.txt
-#
-# Add executables
-add_executable(affinityTest affinityTest.cpp)
-target_link_libraries(affinityTest psimaglite)
-
-add_executable(akimaSpline akimaSpline.cpp)
-target_link_libraries(akimaSpline psimaglite)
-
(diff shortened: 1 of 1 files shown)


g1257:
Note that some (all?) tests may need the following just at the start of `int main()`:

```
constexpr unsigned int nthreads = 1;
PsimagLite::Concurrency(&argc, &argv, nthreads);
```

if it's not there. This is so that MPI_init gets called. I mention this here in case folks pick up work from this MR to extend.

g1257:
All tests should pass now. There's no validation of the results produced though.
I'll leave that for another MR so that they can be converted into catch2 tests or unit tests.
I've also left out tests that need threading enabled with -DUSE_PTHREADS for another MR.

dalg24 (commented):
Please place the `add_test` command right after the declaration of the corresponding executable to improve readability.

g1257:
> Please place the `add_test` command right after the declaration of the corresponding executable to improve readability.

Implemented. Thanks.

PR #75, opened 2025-12-18
More dependency cleanup

dalg24:
Following up on #68 

The PsimagLite handles finding MPI, Boost, as well as BLAS and LAPACK.
DMRG++ get these transitively when linking against psimaglite::psimaglite


g1257 (approved):
Compiles fine; all tests pass on my setup.

PR #82, opened 2026-01-03
Make sure previous run was done with same symmetry as this run

g1257:
Fails if previous run was done with a different symmetry from this run.
We could implement disabling symmetries for the second run, but this isn't done in this PR.

### DIFF
--- a/src/Engine/ParametersDmrgSolver.h
+++ b/src/Engine/ParametersDmrgSolver.h
@@ -687,7 +687,10 @@ struct ParametersDmrgSolver {
 
 		// then for restart
 		bool isRestart = this->options.isSet("restart");
-		return (isRestart) ? this->checkpoint.filename() : "";
+		if (isRestart)
+			return this->checkpoint.filename();
+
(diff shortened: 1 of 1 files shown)


src/Engine/ParametersDmrgSolver.h
```diff
@@ -678,6 +678,18 @@ struct ParametersDmrgSolver {
...
+	PsimagLite::String restartFilename() const
+	{
+		// check first for observe
+		bool isObserve = this->options.isSet("observe");
+		if (isObserve)
+			return this->filename;
+
+		// then for restart
+		bool isRestart = this->options.isSet("restart");
+		return (isRestart) ? this->checkpoint.filename() : "";
```
dalg24: FWIW I think the version in `findHdf5FileIfAny` was more readable with
```suggestion
		if (isRestart)
			return this->checkpoint.filename();

		return "";
```
(not blocking)

dalg24 (approved):

PR #84, opened 2026-01-06
Cincuenta tree and basic cmake for it

PDoakORNL:
I renamed the executable but possible it should just stay cincuenta. Executable is produced but not tested yet.

### DIFF
--- a/cincuenta/src/CMakeLists.txt
+++ b/cincuenta/src/CMakeLists.txt
@@ -2,10 +2,11 @@
 # command and info from dmrgpp/src
 
 add_library(provenance OBJECT Provenance.cpp)
-target_link_libraries(provenance PRIVATE dmrgpp psimaglite)
+target_link_libraries(provenance PUBLIC dmrgpp psimaglite)
 set(dmftpp_src cincuenta.cpp)
 add_executable(dmftpp ${dmftpp_src} $<TARGET_OBJECTS:provenance>)
-target_link_libraries(dmftpp PUBLIC dmrgpp kronutil GSL::gsl)
+target_link_libraries(dmftpp PUBLIC dmrgpp kronutil)
+# GSL::gsl)
 target_compile_definitions(dmftpp PUBLIC USE_GSL)
 
 add_executable(freeGimp freeGimp.cpp $<TARGET_OBJECTS:provenance>)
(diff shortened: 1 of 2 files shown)


g1257:
This is good, but it doesn't run because it needs the GnuScientificLibrary:
```
cd /dmrgpp/build
cincuenta/src/dmftpp -f ../cincuenta/TestSuite/inputs/input.ain 
...
terminate called after throwing an instance of 'PsimagLite::RuntimeError'
  what():  You need to compile with the GSL
```

I suggest adding the GSL as a TPL for cincuenta given that it really needs it.
I wonder if we could add the run above as a test, but it may be too long.
Maybe changing the input.ain so that we take less Matsubaras:
Matsubaras=4;
perhaps it is withing a minute or so.
Else, no worries, we'll add a test in another PR; the numbers aren't reliable anyway.

PDoakORNL:
I'll get this fixed tomorrow.

PDoakORNL:
Still working on this I made the mistake was trying to rebase this onto master, I figured those commits are now in master it will be fine, but maybe because my features that was just not quite up to date with the features that was merged or some more fundamental reason I got an explosion of conflicts. Strangely I got many conflicts in files outside of cincuenta and the files external files I touched in my PR. I'm not sure why it found it ambiguous which commits configure.pl or createGithubRevision.pl it should use, I thought we had a pretty simple master -> features -> master here. Does anyone have an explanation why sometimes rebases like these just blow up? I will look at how messy the merge is now and if that is also a mess I will just build a new PR with this content.

cincuenta/src/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# This is repeated here but should possibly just use the custom
+# command and info from dmrgpp/src
+
+
+
+add_library(provenance OBJECT Provenance.cpp)
+target_link_libraries(provenance PRIVATE dmrgpp psimaglite)
```
dalg24:
```suggestion
target_link_libraries(provenance PRIVATE dmrgpp psimaglite::psimaglite)
```
and we should probably also define an alias and refer to the scoped target for dmrgpp as well (as a follow up work)

PDoakORNL:
Still some formatting to be done for "new" c++ files.

g1257 (commented):
Yes, now running
```
cd build
cincuenta/src/dmftpp -f ../TestSuite/inputs/input.ain
```
runs fine and takes less than 4 seconds in Release mode.

Perhaps keeping the name as ``cincuenta`` instead of ``dmftpp`` is better but it's a suggestion only and not blocking.

PR #86, opened 2026-01-15
Documents that master is now the default branch

g1257:
This PR documents that master is now the default branch, and features is no longer used.
Moreover, we should probably delete features to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,6 @@ cd ../
 git clone https://github.com/g1257/dmrgpp.git
 cd dmrgpp/
 </pre>
-Users should clone and work with the `master` branch.
 Pull requests should be opened against the `master` branch.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,10 +122,9 @@ Create a directory somewhere and cd to it.
 cd ../
 git clone https://github.com/g1257/dmrgpp.git
 cd dmrgpp/
-git checkout features
-git pull origin features
 </pre>
-Please also read https://g1257.github.io/availabilityOfSoftware.html
+Users should clone and work with branch master.
```
dalg24:
```suggestion
Users should clone and work with the `master` branch.
```
I am not sure it is worth saying anything regarding users since that is now the default

PR #94, opened 2026-01-27
Adds the Liouvillian Heisenberg Model

g1257:
This is the first "open system" model example, where the Liouvillian replaces the Hamiltonian and we want the zero eigenvalue state.

Because LiouvillianHeisenberg needs ArnoldiSaI (and cannot use Lanczos or Davidson), some refactoring has been done to the Lanczos/Davidson matrix solvers, so that these two plus ArnolidSaI form a virtual inhertiance hierarchy that is used in Diagonalization.h.

### DIFF
--- a/TestSuite/inputs/input8900.ain
+++ b/TestSuite/inputs/input8900.ain
@@ -1,20 +1,54 @@
 ##Ainur 1.0
 
+# It is highly recommended
+# to read arXiv:2009.08200v1
+# in particular, Eqs~(16), (17) and (18)
+# DMRG++ solvers for a state |x>
+# where L|x> = 0
+#
+# This is the FAT site implementation
+# (a thin site implementation will be forthcoming)
+#
+# This is a sample input file
+# to test a Heisenberg model
+# with jump operators that
+# simulate an open system
+#
+# The example is kept very small
+# so it runs in our CI/CD
+# by default.
+#
+# This is the total size of the chain
+# Each site has 4 states because
+# there are ancillas
+# This is a "fat" site implementation
 TotalNumberOfSites=6;
+
+# There are 4 connections in the Hamiltonian
+# All connectionsa are from i to j where
+# these sites are *different*
 NumberOfTerms=4;
 
+# The Sph^+_iSph-_j term; ph means physical
+# This term is automatically multiplied by -i
 gt0:GeometryKind="chain";
 gt0:GeometryOptions="ConstantValues";
 gt0:dir0:Connectors=[1.0];
 
+# The Sph^z_i Sph^z_j term; ph means physical
+# This term is automatically multiplied by -i
 gt1:GeometryKind="chain";
 gt1:GeometryOptions="ConstantValues";
 gt1:dir0:Connectors=[1.0];
 
+# The Sanc^+_iSanc-_j term; anc means ancilla
+# This term is automatically multiplied by i
 gt2:GeometryKind="chain";
 gt2:GeometryOptions="ConstantValues";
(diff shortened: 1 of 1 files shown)


g1257:
@dalg24 and @PDoakORNL : This is ready for your review. It's somewhat long and there's only one full pipeline test: inpu8900 that you can find in ```src/CMakeLists.txt```

src/Engine/MatrixVectorStored.h
```diff
@@ -129,7 +128,23 @@ template <typename ModelType_> class MatrixVectorStored : public MatrixVectorBas
...
-	SizeType rows() const { return matrixStored_[pointer_].rows(); }
+	SizeType rows() const
+	{
+		assert(pointer_ < matrixStored_.size());
+		return matrixStored_[pointer_].rows();
+	}
+
+	SizeType cols() const
+	{
+		assert(pointer_ < matrixStored_.size());
```
PDoakORNL: the name "pointer_" really makes this confusing to read.  Not really sure what it is supposed to be, sector index?  This class overall needs to some more documentation. 

Why is access is via one state transform and two indexes.
How does the other sector get filled in?

src/Engine/Diagonalization.h
```diff
@@ -97,42 +99,38 @@ template <typename ParametersType, typename TargetingType> class Diagonalization
...
+	typedef typename ModelHelperType::LeftRightSuperType        LeftRightSuperType;
+	typedef typename TargetingType::MatrixVectorType            MatrixVectorType;
+	typedef typename ModelType::InputValidatorType              InputValidatorType;
+	typedef typename PsimagLite::Vector<RealType>::Type         VectorRealType;
+	typedef typename PsimagLite::Vector<SizeType>::Type         VectorSizeType;
+	typedef typename PsimagLite::Vector<VectorRealType>::Type   VectorVectorRealType;
+	typedef PsimagLite::ParametersForSolver<RealType>           ParametersForSolverType;
+	typedef PsimagLite::MatrixSolverBase<MatrixVectorType>      MatrixSolverBaseType;
+	typedef PsimagLite::LanczosSolver<MatrixVectorType>         LanczosSolverType;
+	typedef PsimagLite::ArnoldiSaI<MatrixVectorType>            ArnoldiSaIType;
```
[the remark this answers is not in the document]
g1257: Done for this file.

src/Engine/MatrixVectorStored.h
g1257: It turns out that Lanczos++ doesn't need this file or class. So I went ahead and got rid of the pointer_ thing as suggested, as it's not used in DMRG++.

src/Models/LiouvillianHeisenberg/LiouvillianHeisenberg.hh
```diff
@@ -0,0 +1,490 @@
...
+	{
+		if (!io.doesGroupExist(label1))
+			io.createGroup(label1);
+
+		PsimagLite::String label = label1 + "/" + this->params().model;
+		io.createGroup(label);
+		modelParameters_.write(label, io);
+	}
+
+	void addDiagonalsInNaturalBasis(SparseMatrixType& hmatrix,
```
PDoakORNL: Without a unit test or method documentation I don't really know what to expect this method to do. I'm guessing hmatrix is both in and out not just out.  I don't know what the assumptions about hmatrix are.  I might be able to figure it out if I knew all the methodology and was more familiar with Louvillian's but I'm not.  Some amount of method documentation might be the thing to do at this point.  documentation of the arguments direction and the assumptions about them would help.

PR #95, opened 2026-01-29
support C++ less than 17

g1257:


g1257:
Version 2020 is 1815 pages including the standard library: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/n4849.pdf

PR #96, opened 2026-01-31
Additional tests for the dmrg executable

g1257:
This PR adds additional tests that weren't added before because the input was in the old format. Although DMRG++ still supports the old format, I have converted the inputs to the Ainur format.

If we add somewhat more expensive tests (longer than a minute but not more than 3 minutes long) then we should probably have a flag for that in the cmake maybe. Else we leave for another PR. Suggestions?


dalg24 (approved):
I would introduce the option that enable "heavier" tests in a follow up PR

PR #101, opened 2026-02-05
Typedef into using for the whole codebase

g1257:


PDoakORNL:
So my question @g1257 is do you want this going in before #94 or after? It looks like after to me so I will proceed with that in mind.

g1257:
I prefer #94 before this one. I'll then rebase this one and resolve conflcits if any.

PR #103, opened 2026-02-13
Documents and amends DMRG time evolution 

g1257:
This PR is mostly documentation for the time evolution in DMRG++ by documenting input21.ain and input22.ain.
This PR also fixes the related ```fourierTimeSpace.pl``` script, and fixes missing labels in ```InputCheck.h```. This PR does _not_ address Hamiltonians that are explicitly time-dependent.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -89,7 +89,6 @@ add_test(NAME input10 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input10.ain) # energy 
 # input11 and input12 need -DUSE_PTHREADS
 # add_test(NAME input11 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input11.ain)
 # add_test(NAME input12 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input12.ain)
-<<<<<<< HEAD
 add_test(NAME input20 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input20.ain) # energy -3.37493
 add_test(NAME input21 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input21.ain "<gs|sz|P0>") # none
 add_test(NAME input22 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input22.ain "<gs|sz|P0>") # none


g1257:
This is ready for review. There's going to be another PR about Hamiltonians that are explicitly time-dependent.

TestSuite/inputs/input127.ain
```diff
@@ -21,6 +21,6 @@ TargetElectronsUp=8;
 TargetElectronsDown=8;
 string P0="(2+-i4)*c'[5]*c[6]*|gs>+7.2i*c'[6]*c[7]*|gs>";
 GsWeight=0.2;
-OutputFile="data120";
+OutputFile="data127";
```
dalg24: Avoid unrelated changes

TestSuite/inputs/input21.ain
```diff
@@ -1,36 +1,221 @@
...
 InfiniteLoopKeptStates=60;
-FiniteLoops=[[6, 150, 2],[-6, 150, 2]];
+
+# The finite loops
+# There are two finite loops mentioned explicitly.
+# The first starts from where input20 left off and goes all the way to the left,
+# and the second loop all the way to the right.
+#
+# The value 150 refers to the kept states that will be used.
+# By default a Tolerance is assumed to be 1e-12 so this value of 150
```
dalg24:
```suggestion
# By default the tolerance is 1e-12 so this value of 150
```

TestSuite/inputs/input21.ain
```diff
@@ -1,36 +1,221 @@
...
 RepeatFiniteLoopsTimes=2;
+
+# This is the number of "spins up" for Heisenberg spin=1/2, so we get Sz = 0 as a local symmetry.
+# For spin > 1/2, this is \sum_i (spin + m_i) = const., where the sum over i runs over all sites.
+# Note that in all cases this is a non-negative integer.
+# In other words, DMRG++ uses m + j instead of just m for the symmetry, so that the resulting number
+# is both an integer and non-negative.
 TargetSzPlusConst=4;
 
+# Many vectors will be included in the densiy matrix or actually the SVD decomposition (see paper mentioned above)
```
dalg24:
```suggestion
# Many vectors will be included in the density matrix or actually the SVD decomposition (see paper mentioned above)
```

TestSuite/inputs/input21.ain
```diff
@@ -1,36 +1,221 @@
...
+
+# The delay before stage 2 (operator application), in finite loop units.
+# If g.s. was done in a previous run, the recommended value here is 0 = no delay.
 TSPLoops=[0];
+
+# If more than one operator is mentioned below, this indicates whether they are summed
+# or multiplied. If only one operator is present, then this has no effect.
+#
+# (This method of doing things can only do sums or products, but not
+# sums of products or more complicated things. If those are needed, they should be done
```
dalg24: "can only do sums or products, but not sums of products"
Which is it?

TestSuite/inputs/input21.ain
```diff
@@ -1,36 +1,221 @@
...
+
+# The finite loops
+# There are two finite loops mentioned explicitly.
+# The first starts from where input20 left off and goes all the way to the left,
+# and the second loop all the way to the right.
+#
+# The value 150 refers to the kept states that will be used.
+# By default a Tolerance is assumed to be 1e-12 so this value of 150
+# is the absolute maximum to be used by DMRG++, even if the tolerance
+# wanted could not be reached. The .cout file will produce details.
```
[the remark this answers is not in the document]
g1257: Say user Alice wrote Tolerance=1e-12; and also wrote 150 in the kept states or "m" in the input file. Say that it turns out that with m=150 the error is at best 1e-10. What I meant to say is that the tolerance wanted by Alice will not be reached, because DMRG++ won't do more than 150 states.

Say user Bob wrote Tolerance=1e-12; and also wrote m=1500 (ten times more than before). Say that m=300 is needed to reach an error of 1e-12; that means that the actual "m" used will be 300 tops, not 1500. Here tolerance is reached.

PR #113, opened 2026-02-20
Remove code guarded by USE_SIGNAL

dalg24:
Per discussion yesterday, we do not want to keep it.


g1257 (approved):
I read and pulled and run this one and it looks good.

PR #121, opened 2026-02-27
Optionally logs data that goes to the KronDiag algorithm

g1257:


quantumsteve:
@dalg24 I changed the raw owning pointer to std::optional. Is this ready for review?

dmrg/Engine/MatrixMarket.hh
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+private:
+
+	static std::string buildMMFormatHeader()
+	{
+		return std::string("%%MatrixMarket matrix coordinate real general");
+	}
+
+	const SparseMatrixType& sparse_;
```
quantumsteve, line 95: I don't like these [const reference member types,](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type) but fear making that change could require significant work.

dmrg/Engine/MatrixVectorKron/KronLogger.hh
```diff
@@ -0,0 +1,223 @@
...
+		size_t dot_index = filename.find_last_of(".");
+
+		std::string root = filename.substr(0, dot_index);
+		return "kron_" + root + "_" + ttos(n) + ".txt";
+	}
+
+	static SizeType               counter_;
+	PsimagLite::ProgressIndicator progress_;
+	std::optional<std::ofstream>  fout_;
+	const InitKronType&           init_kron_;
```
quantumsteve, line 218: another [const reference member type](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type)

dmrg/Engine/MatrixMarket.hh
g1257: That's fair and noted in https://github.com/dmrgpp-project/dmrgpp/issues/208
We can do it in a separate PR(s) that deal with the whole codebase maybe.

quantumsteve (approved):

PR #126, opened 2026-03-04
TimeDepConnections: Part III

### LINKED ISSUE #119: Document and amend DMRG time evolution for Hamiltonians that themselves depend on time
Self-explanatory but depends on #117 and #118

g1257:
This PR implements time-dependent connections. For example, for a tight-binding or Hubbard model one could have
```
\sum_{i,j} exp(i*A*t) hop_{i,j} c^\dagger_i c_j + U niup nidown
```
where t in exp(iAt) is physical time.

The time dependence is implemented as a multiplicative factor on top of the regular geometry: GeometryFactor label in the input for that particular connection. This implementation is then model-independent. If the value of A in the above were constant and equal to 42.0 then one could write
```
GeometryFactor=exp:*:1.0i:*:42.0:%t
 ```
Here we use ASTs with colon ```:``` as separator. In the future we could use what I call "canonical expressions," [1] which looks more like normal algebra, or "general expressions" [2] that look almost like normal algebra.
[1] Sort of implemented; [2] not implemented yet: we would here have nested parens; for either see https://github.com/g1257/dmrgpp/blob/master/doc/DmrgPlusPlusFp1.tex

In practive it's better to use a table of value to implement a function ```A(t),``` which is shown in input9001.

If the Hamiltonian has more than one connection (say hoppings and ninj term) as in
```
\sum_{i,j} exp(i*A*t) hop_{i,j} c^\dagger_i c_j + V(t) C(i, j) n_i n_j 
```
then the label ```GeometryFactor``` must be prepended by ``gt`` followed by the index for the "geometry" or connection it belongs to:
```
gt0:GeometryFactor= this would be for the hopping term
gt1:GeometryFactor= this would be for the ninj term or V(t)
```

If GeometryFactor isn't present for a given Hamiltonian connection ("geometry" in DMRG++ jargon) then it is assumed to be 1.0.

Closes https://github.com/g1257/dmrgpp/issues/119

### DIFF
--- a/src/Engine/CookInputExpression.hh
+++ b/src/Engine/CookInputExpression.hh
@@ -1,0 +1,69 @@
+#ifndef COOKINPUTEXPRESSION_HH
+#define COOKINPUTEXPRESSION_HH
+#include "InputCheck.h"
+#include "InputNg.h"
+#include "Matrix.h"
+#include "PsimagLite.h"
+#include <string>
+#include <vector>
+
+namespace Dmrg {
+
+template <typename ComplexOrRealType> class CookInputExpression {
+public:
+
+	using RealType             = typename PsimagLite::Real<ComplexOrRealType>::Type;
+	using InputNgType          = PsimagLite::InputNg<Dmrg::InputCheck>;
+	using InputNgValidatorType = InputNgType::Readable;
+
+	CookInputExpression(const InputNgValidatorType& io)
+	    : io_(io)
+	{ }
+
+	std::string operator()(const std::string& expr)
+	{
+		std::string label = "!readTable";
+		if (expr.substr(0, label.size()) == label) {
+			std::string str = expr.substr(label.size(), std::string::npos);
+			// delete spaces FIXME TODO
+			// split on comma
+			std::vector<std::string> args;
+			PsimagLite::split(args, str, ",");
+			if (args.size() != 2) {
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

File: src/Engine/CookInputExpression.hh [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+#define COOKINPUTEXPRESSION_HH
+#include "InputCheck.h"
+#include "InputNg.h"
+#include "Matrix.h"
+#include "PsimagLite.h"
+#include <string>
+#include <vector>
+
+namespace Dmrg {
+
```
PDoakORNL: If there was every a time for class documentation.

PDoakORNL (changes requested):
Relatively minor changes, but please don't treat const like its optional either use it or don't.

src/Engine/CookInputExpression.hh
g1257: Added docs!

g1257 (commented):
@quantumsteve : I've implemented all your suggestions. Thanks. Let me know what you think now.

g1257:
Friendly reminder to re-review this one.

quantumsteve (approved):

PDoakORNL (approved):
I think this is ready to merge now.

Issue #133, opened 2026-03-05
Brittle hardcoded relative paths must go


dalg24:
These break if the dmrgpp repo is cloned under an alternate name, say `dmrgpp-src` or whatnot (which package managers typically do)
```
cincuenta/src/ImpuritySolverDmrg.h:4:#include "../../dmrgpp/src/Engine/DmrgRunner.h"
cincuenta/src/ImpuritySolverDmrg.h:5:#include "../../dmrgpp/src/Engine/ManyOmegas.h"
cincuenta/src/ImpuritySolverDmrg.h:6:#include "../../dmrgpp/src/Engine/ProcOmegas.h"
cincuenta/src/ImpuritySolverExactDiag.h:4:#include "../../dmrgpp/src/Engine/InputCheck.h"
src/manyOmegas.cpp:2:#include "../../dmrgpp/src/Engine/ProgramGlobals.h"
src/procOmegas.cpp:2:#include "../../dmrgpp/src/Engine/ProgramGlobals.h"
```

These break if we rearrange subdirectories
```
cincuenta/src/InputCheck.h:7:#include "../../PsimagLite/src/Options.h"
cincuenta/src/Provenance.h:3:#include "../../PsimagLite/src/Version.h"
src/Engine/InputCheck.h:80:#include "../../PsimagLite/src/Options.h"
src/Engine/Provenance.h:3:#include "../../PsimagLite/src/Version.h"
src/Models/Graphene/Graphene.h:86:#include "../../../PsimagLite/src/Vector.h"
```

PR #134, opened 2026-03-05
Fix brittle dmrgpp header include with relative paths

dalg24:
Partial fix for #133 addressing what I think was most broken and assuming the name of the source directory to be `dmrgpp/`


PDoakORNL (approved):
Doing these bit by bit seems fine to me.  Easy to review

PR #137, opened 2026-03-07
Enable warnings in CI and treat them as errors

dalg24:
I enabled `-Wall -Wextra` and had to disable
* `-Wignored-qualifiers`
* `-Wunused-local-typedef`
* `-Wunused-parameter`
* `-Wunused-variable`
* `-Wno-sign-compare`
We can attempt to fix these in follow up as appropriate but I think we should put this in place to avoid introducing more.


g1257:
@dalg24  : I commited to your branch and fixed everything except for unused-parameter, so we keep only -Wno-unused-parameter

dalg24:
Do you know where it is emitted from?
If so, we could temporarily disable that diagnostic.
```C++
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Warray-bounds"
// Code with warning
#pragma GCC diagnostic pop

g1257:
@dalg24  :  If you want to delete sumDecomposition.{h,cpp} and the PsimagLite/examples/CmakeLists.txt associated line, please do so. This isn't used anywhere anymore apparently.
(If we had enough time, I'd wonder what mistake I'm making in that C++ code that the compiler doesn't like. I mean, some day I hope to learn to program in C++)

PR #148, opened 2026-03-19
Suppress HDF5 warnings in input2 test

quantumsteve:
This adds a `HDF5DisableExceptionPrinting` class for temporarily suppressing HDF5 warnings and adds it where these warnings can be ignored.

I also did a bit of modernization to remove raw owning pointers.

This is part of #128


g1257:
Ah, never mind about my comment on reviewers, I see you added me, that's good.

g1257 (approved):

PR #149, opened 2026-03-19
Lanczos++ added to repo.

PDoakORNL:
I'm at a impass here in that Lanczos++ seems incompatible with the API of the Geometry "module" of the current version of PsimagLite in dmrgpp.  @g1257 I don't want to modify anything in Geometry as there seem to be two pretty complex ideas of how to present the Geometry to consumers, they seems like they may be parallel but `Geometry` seems to have been shedding API elements that Lanczos needs, while new development seems focused on the GeometryBase and Geomtetry Subtypes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,17 @@ if(BUILD_TESTING)
   include(Catch)
 endif()
 
+find_program(AWK NAMES awk mawk gawk)
+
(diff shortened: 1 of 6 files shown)


g1257:
I'm not blocked, no, I can proceed with the coding I need to do regardless. I mean, at least for while.

PR #169, opened 2026-04-08
Cincuenta fix exact solver

### LINKED ISSUE #109: Cincuenta: Exact solver and DMRG Solver Differ too much
The title explains. This is a bug but I don't know which solver is correct, if any.

### LINKED ISSUE #165: Change name of dmrg library
The dmrg library in src/CMakeLists.txt is really a few linkage symbols that are coming from ProgramGlobals.o and related but nothing core.
```
add_library(dmrgpp ProgramGlobals.cpp Provenance.cpp Utils.cpp Su2Related.cpp Qn.cpp)
```
I propose to change the name then to dmrgpp_utils or something like that. 

Motivation: The dmrg_runner will be a DMRG library including core stuff, so we need to distinguish it from utils.

g1257:
This PR does the following things.

(1) It creates lanczos_globals as a library so that cincuenta can use Lanczos++ "package" with all symbols from it in LanczosGlobals.cpp.

(2) cincuenta now uses Lanczos++ for the exact solver so that a lot of its code is deleted.

(3) Creates a new cincuenta test so that we have now to mostly identical tests: one with exact solver and one with dmrg solver: they yield the same results now.

Closes https://github.com/g1257/dmrgpp/issues/109
Closes https://github.com/g1257/dmrgpp/issues/165


cincuenta/src/ImpuritySolverDmrg.h
```diff
@@ -6,7 +6,6 @@
 #include "Geometry/Star.h"
```
g1257: Geometry/Star.h is in PsimagLite/src

PDoakORNL (approved):
So here's my take on this, this leads to more correct EQ DMFT and matching ED and DMRG results.  I managed to understand it well enough to work with agents to add initial Non-equilibrium DMFT using ED to it and sensible results came out. I don't know cincuenta well enough yet to comment on whether the refactoring is perfect or not but it looks reasonable. I don't see any crimes against software engineering being made here so I think this should get merged.

g1257:
> So here's my take on this, this leads to more correct EQ DMFT and matching ED and DMRG results. I managed to understand it well enough to work with agents to add initial Non-equilibrium DMFT using ED to it and sensible results came out. I don't know cincuenta well enough yet to comment on whether the refactoring is perfect or not but it looks reasonable. I don't see any crimes against software engineering being made here so I think this should get merged.

Thanks for the feedback but remember that this is still in draft because I'm breaking it down in parts: I've just merged modular_lanczospp that you reviewed (thanks!) and the diff now is smaller between this PR and master. I will now start creating another PR that includes changes in this PR that touch PsimagLite as another smaller part.

PR #173, opened 2026-04-09
CMake cleanup following the integration of Lanczos++

dalg24:
Most importantly installing the new executable

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,9 +41,5 @@ endif()
 add_subdirectory(PsimagLite)
 add_subdirectory(src)
 
-add_subdirectory(cincuenta)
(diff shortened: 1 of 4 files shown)


g1257 (approved):
This is fine for now, but in the future we'll need to clean up statics in LanczosPlusPlus directory and provide a lanczos_global library with symbols needed by things that want to use LanczosPlusPlus (like cincuenta in the PR I'm working on)

That covers the linkage part. The headers part would have to be defined but we don't do that for now because we don't have an appropriate API for any of the "packages": dmrgpp, PsimagLite, LanczosPlusPlus. I believe cincuenta will be a "final" package not used by anybody, but we could add it to the list.

PR #174, opened 2026-04-10
introspect fixed and test added

g1257:
Fixes introspect.cpp and adds a test that takes virtually no time. This PR takes precedence over https://github.com/g1257/dmrgpp/pull/175


PDoakORNL (approved):
This looks fine to me.  And it did work for me when I tried it.

PR #175, opened 2026-04-10
Input path

g1257:
This PR will be merged AFTER https://github.com/g1257/dmrgpp/pull/174

Closes https://github.com/g1257/dmrgpp/issues/170


PsimagLite/src/InputPath.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace PsimagLite {
+
+class InputPath {
+public:
+
+	InputPath()
+	    : paths_(1, "")
+	{ }
+
+	void push(const std::string& path)
```
dalg24: Should we be storing `std::filesystem::path` instead of strings?

PsimagLite/examples/inputPath.cpp
```diff
@@ -0,0 +1,32 @@
...
+			throw std::runtime_error(usage);
+		}
+	}
+
+	if (filename.empty()) {
+		throw std::runtime_error("File must be provided\n");
+	}
+
+	std::string result = input_path.findFirst(filename);
+	std::cout << result << "\n";
```
dalg24: It would be better to use the Catch2 unit test framework to test this functionality.

PsimagLite/src/InputPath.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <string>
+#include <vector>
+
+namespace PsimagLite {
+
+class InputPath {
+public:
+
+	InputPath()
+	    : paths_(1, "")
```
dalg24: Add a comment that the working directory takes precedence.  I actually find that counterintuitive.
I think one possible way to fix it would be to iterate in reverse order, which incidentally means that the last cmd line arguments wins.

cincuenta/src/cincuenta.cpp
```diff
@@ -151,14 +152,21 @@ to the main dmrg driver are the following.
...
+	if (precision > 0) {
 		params.precision = precision;
+	}
+
 	params.echoInput = echoInput;
 
+	params.gsTemplate    = input_path.findFirst(params.gsTemplate);
+	params.omegaTemplate = input_path.findFirst(params.omegaTemplate);
+
+	// END adjust params
```
dalg24, line 169: Did you mean to commit this?  Is that a drive-by change?

g1257: 166 and 167  are crucial. 

I guess you mean the rest of the lines there I just reorganized adjustment to params in a better way. I mean, the adjustment that happen after we read them in line 158.

dalg24: > 166 and 167 are crucial.

OK

PsimagLite/examples/inputPath.cpp
g1257: OK. I just tried this for like an hour but can't find a good way to pass -I ${CMAKE_SOURCE_DIR} to the catch2 test. I mean, that's the whole purpose of the test.
I have followed the tutorial and write my own int main() and added the option which I guess I can pass with EXTRA_ARGS. However, catch2 executes the test during build (!) which should not happen. Any ideas?

PsimagLite/src/InputPath.hpp
g1257: Yes, agreed. Fixed.

PsimagLite/src/InputPath.hpp
g1257: I'd say no. The string manipulation is what's important. std::filesystem:path is briefly used to check file existence (we should check readability too but that'd be off topic here).

dalg24 (approved):
@g1257 go ahead and merge if you are ok with my changes to InputPath and the unit test I added

g1257:
@dalg24 : Looks great! May I merge now?

dalg24:
> @dalg24 : Looks great! May I merge now?

Go ahead

PR #179, opened 2026-04-23
Add GPUPlugin to dmrgpp

quantumsteve:
This adds and builds [DmrgppPluginSc](https://github.com/g1257/DmrgppPluginSc). It also builds and runs two provided tests, neither of which appear to check correctness.

Yesterday we decided to add this now with minimal changes so future pull requests can be smaller and easier to review.


g1257:
@dalg24 : If you have time to have a quick look at the cmake lists in this PR so that we can approve and merge if there are not objections. Else we can also ask @PDoakORNL . 

 @quantumsteve  can then do follow-up work in other PRs,  when this things is brought into the repo.

g1257 (approved):

PR #182, opened 2026-05-12
Fix undefined symbol `dmrg_free<void>(void*) -> void`

dalg24:
```
Undefined symbols for architecture arm64:
  "void dmrg_free<void>(void*)", referenced from:
      unsetup_vbatch(int**, int**, int**, double**, double**) in unsetup_vbatch.cpp.o
      unsetup_vbatch(int**, int**, int**, double**, double**) in unsetup_vbatch.cpp.o
```


PDoakORNL (approved):
LGTM

PR #185, opened 2026-05-13
Increase CI coverage (build types, shared libs)

dalg24:
Going from a single CI build to 3 with arbitrary combination of options.


dalg24:
All the failures in the build with Boost disabled are improper guarding of the tests
```
terminate called after throwing an instance of 'PsimagLite::RuntimeError'
  what():  To use Ainur, you need boost-devel, and compile with -DUSE_BOOST
```

If we confirm we want to support disabling Boost, we can go through and guard them.
@g1257 how do you want to handle this?

g1257:
Why are we running with BOOST disabled? I thought we decided it was mandatory.

Also, how should one properly guard the tests?
Should one do in cmake something like
if (TESTS_ON and BOOST_ON) 
    run_tests
endif
or something like that?

dalg24:
Currently Boost is an optional dependency, that's why I explicitly disabled it in one of the CI builds.
If we are wanting to make it a required dependency, then I can add the REQUIRED keyword to find_package() and we don't need to test without Boost.
If instead we do want to support Boost as optional dependency, then yes we need to do something along the lines of what you said.

Let me know what you want to do.

g1257:
Please make BOOST mandatory: BOOST is used for Ainur, the input parser for the new input format, which we prefer over the deprecated POD input format. (We may want to remove support for the POD inputs, but that would be a discussion for another day.)

Thanks.

quantumsteve:
Let's coordinate with #188

g1257 (approved):

PR #186, opened 2026-05-13
Performance Inputs in CI

g1257:
@dalg24 : Here are the performance inputs that I had sent you before; the seem to pass the CI; input340.ain, input341.ain, and input342.ain, but input343.ain takes too long so it's disabled.


g1257:
@dalg24  : This could be merged at some point if you it worth. Perhaps we'll need to define a HEAVY_TESTS_ENABLE or so variable so that the CI doesn't run them normally, only occasionally. Please advise.

masterleinad:
I think the run time is fine if we exclude these tests for `Debug` builds.

masterleinad:
I'm currently using these inout files for benchmarking so it would be helpful if we could merge them in some sort or fashion.

g1257:
@masterleinad : Rebased off of master now. If you want to protect these tests (340, 341, 342) so they don't run in debug mode or other condition let me know how, or feel free to push changes directly to this branch. After that we can move to ready for review and merge.

g1257:
@masterleinad  : This is ready for your review.

masterleinad:
We will need https://github.com/dmrgpp-project/dmrgpp/pull/220 to ensure these tests are actually skipped in `Debug` mode.

masterleinad (approved):

PR #190, opened 2026-05-28
Make Boost a required dependency

dalg24:
Per https://github.com/dmrgpp-project/dmrgpp/pull/185#issuecomment-4454906960
The project did not build when Boost was not found.  We decided to make it a required dependency rather than fixing, which makes our lives a bit easier to test and maintain.


dalg24:
Noting that the README was actually saying Boost is required
https://github.com/dmrgpp-project/dmrgpp/blob/a6c4b986a39992621ddf9181388731e25362b037/README.md?plain=1#L109-L110

g1257 (approved):

PR #193, opened 2026-05-29
scaling gimp by 0.25

g1257:
The Gimp scale of DMRG is correct and should not be scaled. The scale of EXACT is wrong and this PR fixes that temporarily. We'll need to fix properly later.

### DIFF
--- a/cincuenta/src/ImpuritySolverBase.h
+++ b/cincuenta/src/ImpuritySolverBase.h
@@ -64,6 +64,14 @@ public:
 
 		return sum;
 	}
+
+	static void scale(std::vector<ComplexOrRealType>& g)
+	{
+		const ComplexOrRealType factor = 0.25;
+		for (auto& val : g) {
+			val *= factor;
+		}
+	}
 	// public static END
 
 protected:
--- a/cincuenta/src/ImpuritySolverDmrg.h
+++ b/cincuenta/src/ImpuritySolverDmrg.h
@@ -81,7 +81,7 @@ public:
 
 		doType(DmrgType::TYPE_1, data3, impurity_site, freq_enum);
 
-		// scaleGimp();
+		BaseType::scale(gimp_);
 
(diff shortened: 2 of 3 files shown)


cincuenta/src/ImpuritySolverBase.h
```diff
@@ -64,6 +64,14 @@ template <typename ComplexOrRealType> class ImpuritySolverBase {
 
 		return sum;
 	}
+
+	static void scale(std::vector<ComplexOrRealType>& g)
```
PDoakORNL, line 68: For now this needs a very intense warning/explanatory comment and a TODO to remove when the conventions for what the scaling of G_{imp} are established.  

I would like to think, maybe naively that we could have both a G_{imp} and spectral functions (A) that are scaled by a convention that is clear.  Any factors due to sectors included, semicircular vs. ? DOS etc are easily discoverable in the docs and even at runtime.

PDoakORNL (dismissed):
I'd like this in soon because it seems better this be here than in the cincuenta code farther from its origin, but I'd very much like to understand the why of this scaling long term and how the code can communicate its scaling convention to other components of the code.

cincuenta/src/ImpuritySolverBase.h
g1257: It appears that this isn't by any means settled.

PDoakORNL:
My feeling on this has changed I don't think this is the right way to do this.  The point at which a sum rule or some rigorous way of normalizing should be used.

g1257:
@PDoakORNL : This PR scales only the exact diag solver by 1/4. This is a temporary measure until I fix the real factor problem in Lanczos++'s Engine's calcSpectral(). If you think it makes sense for others to have for now until the real fix, go ahead and approve and merge, but if not it's fine too. We'll merge the actual fix when ready.

g1257:
Addressed somewhere else but still needs to divide by 4.

PR #195, opened 2026-06-03
fixing treatment of bath sites for half filling

PDoakORNL:
Claude Sonnet 5.6 "coauthored" this code.

Tests for this are highly entangled with the non-eq DMFT code so it will be more expedient for the testing to catch up with this change in later PR's


g1257 (approved):
The description should say: "Bug fix for the derivative of the Anderson function with respect to Vs and onsite energies, in the case when particleholesymmetric option is active."

Other than that, it's good to go. Approved.

g1257:
@PDoakORNL : Feel free to merge this PR if you want, so others working on cincuenta can benefit from this bug fix. Up to you.

PR #197, opened 2026-06-04
add TimeEvolve overlap=1 for consecutive P-vector overlap output

PDoakORNL:
Physics motivation: tDMRG-based neq-DMFT requires the time-evolved pre-quench GS bra
|Ψ_0(t)> = exp(-iH_f t)|GS_i> for correct G^>(t,0) measurements with a post-quench
Hamiltonian H_f ≠ H_i.

New feature: overlap=1 option on TimeEvolve{} strings in TargetingExpression.
```
string P2="TimeEvolve{tau=0.1,steps=2,advanceEach=1,overlap=1}*|gs>";
```
Outputs `<Pn|prev>` to the in-situ log after each advance — the overlap of the P-vector
with its state from the previous time step. This consecutive N-sector overlap is used
for gauge-phase correction when measuring cross-sector quantities like `<P2|c|P1>`.

## Test scheme (9100 → 9102 → 9103)

- **9100**: GS under H_i (U=0)
- **9102**: TargetingExpression restart, computes P0 = c'[0]*|GS_i> under H_f (U=2)
- **9103**: TimeEvolve with overlap=1, measures `<P2|c|P1>` = G^>(t,0) in-situ; `<P2|prev>` output verified in log


PDoakORNL:
Consider this to be a work in progress again. The CI reveals some sub par design going on with targetVectors_ which I'm straightening out.

PDoakORNL:
So one thing that is getting done here that is in my opinion not great is targetVectors_ is getting semantically expanded to contain more than the Krylov target vectors.

To paraphrase Claude on this:
The GS evolution slot ends up here rather than in a separate member due to the measurement plumbing.
TargetingCommon::getVector() resolves "P" braket labels by indexing directly into targetVectors_.
For cocoon() to evaluate <gsT|c|P1> it needs evolvedGsIdx_ to be a valid index into that same array — there's no other code path for registering an  arbitrary VectorWithOffsetType* as a named measurement target.

This is the sort of cutesy stuff you see in code written to get the next paper all the time, and I imagine this is where Claude picked it up as well. If you can let it slide @g1257 I do have a plan for how it can be refactored into something sane, but I'm trying to keep the complexity of getting the first cut of the neq-DMFT in.

g1257:
In 9100 you use H0 to obtain its gs, call it |gs>.
In 9101 you use H1 to compute <gs|exp(iH1 t) c_i exp(-iH1 t) c_0|gs>, where |gs> is from before.
If this is the case, we don't need new code, I don't think. We can use TargetingExpression instead of TimeStepTargeting and go
9100 (as before) -->  9102 (see below) --> 9103 (see below)

9100 is unchanged and what you have.

Do not use 9101.

# 9102 is below:
TotalNumberOfSites=6;
NumberOfTerms=1;
DegreesOfFreedom=1;
GeometryKind=star;
GeometryOptions=none;
hubbardU=[2, 0., 0., 0., 0., 0.];
Model=HubbardOneBand;
SolverOptions=twositedmrg,geometryallinsystem,TargetingExpression,restart;
Version=TSPEvolveGroundState-unit-test;
OutputFile=data9102;
InfiniteLoopKeptStates=50;
FiniteLoops=[
    [@auto, 50, 2], [@auto, 50, 2]];
TargetElectronsUp=3;
TargetElectronsDown=3;
dir0:Connectors=[0.5, 0.5, 0.5, 0.5, 0.5];
potentialV=[0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0];
RestartFilename=data9100;
string P0="c'[0]*|gs>";
GsWeight=0.1;

Here we just compute P0 as defined there. Note the 2 in the triplets to avoid recomputing the g.s.

Now we go from 9102 --> 9103 where 
# 9103 below:
TotalNumberOfSites=6;
NumberOfTerms=1;
DegreesOfFreedom=1;
GeometryKind=star;
GeometryOptions=none;
hubbardU=[2, 0., 0., 0., 0., 0.];
Model=HubbardOneBand;
SolverOptions=twositedmrg,geometryallinsystem,TargetingExpression,restart,usecomplex;
Version=TSPEvolveGroundState-unit-test;
OutputFile=data9103;
InfiniteLoopKeptStates=50;
FiniteLoops=[
    [@auto, 50, 2], [@auto, 50, 2],
    [@auto, 50, 2], [@auto, 50, 2],
    [@auto, 50, 2], [@auto, 50, 2]];
TargetElectronsUp=3;
TargetElectronsDown=3;
dir0:Connectors=[0.5, 0.5, 0.5, 0.5, 0.5];
potentialV=[0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0];
RestartFilename=data9102;
RestartMappingTvs=[0, -1, -1];
string P0=|P0>;
string P1="TimeEvolve{tau=0.1,steps=5,advanceEach=4}*|P0>";
string P2="TimeEvolve{tau=0.1,steps=5,advanceEach=4}*|gs>";
GsWeight=0.1;

where P0 is as before, and P1 and P2 should be clear. We would then measure 
<P2|c|P1> in situ.

Again, this doesn't need any code change from master.

PDoakORNL:
I'm pretty impressed this is possible with P vectors, this didn't occur to me as a solution.  I'm looking at whether this can meet all the needs that this change met.  There is one need that the tests didn't capture that is still an open question.

PDoakORNL:
Ok in light of useful PR feedback we've removed the semantic smearing of targetVectors_ and are making better use of DMRG's flexible evaluation abilities.  There is still the need add one small new feature to allow the evaluation of a p-vector with its previous time step.

PDoakORNL:
@g1257 Ready for a new review

dmrg/CMakeLists.txt
```diff
@@ -185,6 +185,21 @@ add_test(NAME input9000 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input9000.ain) # ene
...
 
+# U=0->U=2 quench unit tests using TargetingExpression + TimeEvolve overlap tracking
+# input9100: GS of H_i (U=0); energy = -sqrt(5) for uniform hopping 0.5
+add_test(NAME input9100 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input9100.inp) # energy -2.23607
+add_lowest_eigenvalue_check_test(output9100 -2.23607 runForinput9100.cout)
+# input9102: TargetingExpression restart from 9100, computes P0 = c'[0]*|GS_i>
+add_test(NAME input9102 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input9102.inp)
+set_tests_properties(input9102 PROPERTIES DEPENDS input9100)
+# input9103: tDMRG under H_f, P2=TimeEvolve{overlap=1}*|gs>, measures <P2|c|P1>
+add_test(NAME input9103 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input9103.inp "<P2|c|P1>")
```
g1257, line 196: Instead of adding the overlap feature, is it possible to add the measurement here?
Something like
```
add_test(NAME input9103 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input9103.inp "<P2|c|P1>,<P9|c|P1>")
```
or something like that. Not sure what exactly is the measurement you want.

PDoakORNL: The `<P2|prev>` output isn't a matrix element — it's the consecutive time-step self-overlap `<P2(t) | P2(t−τ)>`. We need it for a gauge-phase correction in the neq-DMFT driver.

In the neq-DMFT context, P1 lives in the N+1 particle sector and P2 in the N sector. Their relative phase drifts sweep-by-sweep because DMRG applies independent unitary gauge choices to each sector. To reconstruct G^< correctly, we accumulate the product of these consecutive overlaps to track and cancel that drift. A standard in-situ operator measurement like `<P2|c|P0>` is a fixed-time matrix element — it has no memory of P2 at the previous time step — so it can't substitute here.

The facility is small: one boolean added to `TimeParams` and roughly 6 lines of output code in `timeEvolve()`. Happy to discuss the gauge-correction formulation further if that would help.

PDoakORNL: Thanks for the guidance on `steps` and `advanceEach` — updated input9103 to `steps=5, advanceEach=4` (N−2 for the 6-site chain). Tests still pass.

We also looked into the auto-generated P-label approach you described. You're right that `indices[last]` holds exp(−iHτ)|ψ(t)⟩ during the non-advance sweeps, so `⟨P_last|P2⟩` would give the consecutive overlap. Our hesitation is that `cincuenta` generates the DMRG++ input programmatically, and the label for `indices[last]` is the `targetVectors_` slot index at allocation time — a number that depends on how many vectors were allocated before it. We're not sure how fragile that is in practice: would adding another P-vector expression shift the index? Would it be stable across DMRG++ versions?

If you think the slot index is reliable enough, or if there's a clean way to query it, we're happy to drop `overlap=1` entirely. We just want to make sure `cincuenta` isn't depending on an internal detail that could silently change.

g1257: Maybe something like 
```
 ./dmrg -f /path/here/input9103.ain  "<P2|c|P1>,<P1.last|P1>"
```
is robust enough, so that we are sure that P1.last is the last vector in the series built for the P1 time evolve.
We can discuss the exact syntax (".last"); it would have to be implemented though.

PDoakORNL: That seems like it would work.  Is there any label syntax with `.something` already?

g1257: No, there isn't.

PDoakORNL: It seems less obscure than adding an attribute to the TimeEvolve tag. I think its an advantage that "<P1.last|P1>" looks like an overlap. Do you mind if I take a try at implementing it?

PDoakORNL:
Since we arrived at a better way to do this, this PR superceded by #202

PR #200, opened 2026-06-08
test: add U=0 PH-symmetric G_imp=G_cluster0 regression test

PDoakORNL:
I feel like this gives us a relatively bullet proof wayt to use ED solver from cincuenta's  DMFTSolver.  It's my understanding that normalization from fitting the high frequency Matsubara tail is a clean way to normalize G_imp since we have access to I think arbitrarily high Matsubara frequencies.  This will give us something we can safely do self consistency with.

Ok so the robot assisted PR description:

At U=0 the self-energy is identically zero, so the impurity solver must reproduce the non-interacting cluster propagator G_cluster0 exactly. A bath-parameter layout bug (fixed in PR #195) and a missing spectral normalization in ImpuritySolverExactDiag caused G_imp ~= 4x G_cluster0 with FitOptions=particleholesymmetric.

Fix: add enforceSpectralSumRule() to ImpuritySolverBase, which rescales gimp after each Matsubara solve using the high-frequency tail criterion -wn*Im[G(iwn)] -> 1. ImpuritySolverExactDiag implements scaleGimp(); ImpuritySolverDmrg opts out (has its own normalization path).

DmftSolver::writeGcluster0ForDebug() writes gcluster0_<solver>.txt for direct comparison against gimp_<solver>.txt in the test. The new test u0_gImpEqGcluster0_exactdiag runs with inputU0ExactDiag.ain (U=0, FitOptions=particleholesymmetric, DmftIter=1) and checks max|G_imp - G_cluster0| < 1e-3 (measured residual ~2.3e-4, 4x headroom).


PDoakORNL:
For now I think incoming cincuenta PR's test this sufficiently.

PR #209, opened 2026-06-24
Ensure add_lowest_eigenvalue_check_test starts after corresponding test has run

masterleinad:
Currently, there is no dependency between `add_test` and `add_lowest_eigenvalue_check_test` so that spurious test failurs are observed when running `ctest` with multiple processes. This pull request adds a depednency using `DEPENDS`.
I decided to be explciit about the dependent target instead of using regexes to deduce it inside if `add_lowest_eigenvalue_check_test`.


PDoakORNL (approved):
This is how this should have been done, we shouldn't actually depend on the tests running serially.  Everyone lookout for this issue elsewhere I know i have some cincuenta tests in recent work that this applies to and I will get them fixed.

PR #212, opened 2026-06-26
Ainur: Unicode chars allowed

### LINKED ISSUE #196: Ainur chokes on unicode in comments
Since we are very often using equations that can be expressed in unicode  i.e. greek letters, subscripts, etc it would be nice if Ainur would except such in the comments of input files.  It would also be less implicitly anglocentric which ascii certainly is. Is this a bug, I consider it one.  Programs in 2026 shouldn't have a fatal exception if they get an UTF-8 encoded file.

g1257:
Closes https://github.com/dmrgpp-project/dmrgpp/issues/196

### DIFF
--- a/TestSuite/inputs/input20.ain
+++ b/TestSuite/inputs/input20.ain
@@ -1,5 +1,8 @@
 ##Ainur 1.0
 
+# α is the first letter
+# β is the second letter
+#
 TotalNumberOfSites=8;
 NumberOfTerms=2;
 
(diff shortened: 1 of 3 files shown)


TestSuite/inputs/input20.ain
```diff
@@ -1,5 +1,8 @@
 ##Ainur 1.0
 
+# α is the first letter
```
PDoakORNL: Is this literary or does this explain something about Ainur's operation now that unicode characters are accepted?

PDoakORNL (approved):
Thanks @g1257

TestSuite/inputs/input20.ain
g1257: Moved now to PsimagLite/src/Ainur/input0.ain and directly mention in the comments that we are testing for unicode acceptance.

PR #215, opened 2026-07-07
Make RelWithDebInfo the default CMake build type

masterleinad:
Providing a default build type makes it much easier to get a sensible build by default. Previously, if no `CMAKE_BUILD_TYPE` was provided `CMake` would not add any debugging or optimization flags.


CMakeLists.txt
```diff
@@ -5,6 +5,14 @@ project(
   VERSION 1.1
   LANGUAGES CXX)
 
+# Set default to RelWithDebInfo if not specified
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  set(CMAKE_BUILD_TYPE
+      "RelWithDebInfo"
+      CACHE STRING "Choose the type of build" FORCE)
```
dalg24, line 12:
```suggestion
      CACHE STRING "Default build type" FORCE)
```

dalg24 (approved):

CMakeLists.txt
masterleinad: It's not just the default build type. You can still change it after first configuring.

PR #216, opened 2026-07-08
Use Kokkos for csr_matmul

masterleinad:
The crs_matmul kernels appear to also be hotspots so we tackle them next.


masterleinad:
Not clear if this is relevant for now.

PR #217, opened 2026-07-08
Add tDMRG-based non-equilibrium impurity solver for cincuenta

PDoakORNL:
## Summary

- Adds `ImpuritySolverNeqTdmrg`, a non-equilibrium impurity solver selected via `NeqSolver="tdmrg"` that uses DMRG++'s `TargetingExpression`-based tDMRG (separate particle- and hole-sector runs) to compute the neq-DMFT impurity Green's functions, as an alternative to exact diagonalization.
- Uses the same fixed, equilibrium-fitted bath as `ImpuritySolverNeqExactDiag` (no bath modification during the quench) — this PR validates tDMRG as a correct alternate solution method for that model, not a physics improvement over ED. See scope note below.
- Adds `compare_neq_gr.py` for comparing tDMRG vs. exact-diag G^R/G^< output.

## Tests

- `neqTsujiTdmrg0` / `neqTsujiExactDiag0` / `neqTsujiTdmrg0_compare`: U_i=0->U_f=2 Tsuji quench benchmark; tDMRG vs. exact-diag agree within the tolerances used in the earlier gsT-engine campaign (mean |dRe| ~ 0.009, mean |dIm| ~ 0.007 for m=200, dt=0.1, 6 sites).
- `u0_neq_gImpEqGcluster_tdmrg(_check)`: U=0 plumbing/pipeline check. Transitive argument: `u0_neq_gImpEqG0_exactdiag_check` already proves ExactDiag G_imp = G_0 at U=0; this proves tDMRG G_imp = ExactDiag G_imp, so tDMRG G_imp = G_0 by transitivity. Not an accuracy test — U=0 has no entanglement growth, so this is trivially satisfied.
- Full local build + `ctest -E Nightly`: all cincuenta/tDMRG tests pass; 10 pre-existing, unrelated DMRG++ core test failures observed (introspect0_0, input121/122/123/127, input3021, input8901, testIoNgBoolean0) — untouched by this diff, present on master already.

## Scope note

This solver currently uses only the equilibrium-fitted "first" bath — same as `ImpuritySolverNeqExactDiag` — and does not yet include GBEK's second-bath (rank-L Cholesky) construction that represents the non-equilibrium correlation piece (Delta+) of the hybridization (see PRB 88, 235106, Sec. IV.C). Consequently tDMRG agrees with static-bath ExactDiag by construction (same Hamiltonian, different solver), not with GBEK. Adding the second bath to the tDMRG chain is tracked as follow-on work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cincuenta/src/cincuenta.cpp
```diff
@@ -187,10 +188,27 @@ int main(int argc, char** argv)
 			std::cout << "\n=== Non-equilibrium DMFT (interaction quench) ===\n";
 			ParamsNeqType neqParams(io);
 
-			using ExactNeqSolverType = Dmft::NeqDmftSolver<std::complex<RealType>>;
-			ExactNeqSolverType neqSolver(neqParams, io);
-			neqSolver.solve(dmftSolver.bathResult());
-			neqSolver.dumpGreenFunctions();
+			// Check for tDMRG solver selection
```
g1257, line 191: This block at the end of cincuenta.cpp is starting to grow. I'd leave it like that for now but eventually we may want to think if it should go into DmrgSolver or somewhere else.

So, just for my understanding, the logic overall goes like this: equilibrium is always done as the first part (before this block) and then this thing is done if isNeq (= tMax > 0), correct?

cincuenta/src/ImpuritySolverNeqTdmrg.h
```diff
@@ -0,0 +1,644 @@
...
+			potGS[i + 1]    = bathEps[i];
+			potTdmrg[i + 1] = bathEps[i];
+		}
+
+		// Run 1: GS DMRG for H(U_i)
+		std::cout << "ImpuritySolverNeqTdmrg: GS DMRG U_i=" << params_.uInitial << "\n";
+		{
+			Dmrg::CmdLineOptions opts;
+			opts.logfile = root_ + "gs.log";
+			DmrgRunnerType runner(
```
g1257, line 132: I'm so happy that all entries into DMRG core are done through the DmrgRunner. In retrospect it was a good thing that we gated all of it this way and seems to work even for neq. Of course, the runner can change in the future but at least there's only one gate and no side doors to the dmrg itself. Funny that the interface is a string (the input "file").

cincuenta/TestSuite/inputs/inputNeqTsujiTdmrg.ain
```diff
@@ -0,0 +1,97 @@
...
+
+NeqDmftIter=1;
+NeqDmftTolerance=0.001;
+
+# ---- tDMRG impurity solver selection ---------------------------------------
+NeqSolver="tdmrg";
+
+# tDMRG finite loops: TSPAdvanceEach=4 gives ~1 advance per sweep.
+# For NtNeq=50: 60 rows covers 50 advances with 10 extra at t_max.
+matrix FiniteLoopsTdmrg=[
```
g1257, line 61: I mention this in passing even though you may not be able to use it directly.
In a regular input file, the input line
```
## Example 1
FiniteLoops=[[a,b,c],[d,e,f], [a,b,c],[d,e,f], [a,b,c],[d,e,f]];
```
is equivalent to
```
## Example 1 (short version)
FiniteLoops=[[a,b,c],[d,e,f]];
FiniteLoopsTimes=2;
```
That is, whatever is in FiniteLoops is first written "2" times, and then _appended to the end_ of whatever was there in the first place.

Another example is this
```
## Example 2
FiniteLoops=[[a,b,c],[d,e,f]];
FiniteLoopsTimes=2;
FiniteLoopsFrom=1; # in units of triplets
```
where whatever is in finite loops starting from triplet number "1" to the end, is copied "2" times (and we get [d,e,f], [d,e,f], and then added to whatever was in finiteloops in the beginning. So the above is lowered to
```
## Example 2 (lowered)
FiniteLoops=[[a,b,c],[d,e,f] [d,e,f], [d,e,f]];
```
You can also set the end point, and the below example 3 lowers to yield the same as Example 2.
```
FiniteLoops=[[a,b,c],[d,e,f],[d,e,f]];
FiniteLoopsTimes=1;
FiniteLoopsFrom=1; # in units of triplets
FiniteLoopsTo=2; # in units of triplets, does not include this triplet
```
Notice that the original finite loops is always appended to.

g1257 (approved):
As you can see my review consists of comments only; no requests; so approved, and merge when you want.

cincuenta/src/ImpuritySolverNeqTdmrg.h
PDoakORNL: It's working well so far and it is in own way an API.  I think as long as we are handing off large pieces of work to DMRG we don't need to mess with it.

cincuenta/TestSuite/inputs/inputNeqTsujiTdmrg.ain
PDoakORNL: I can see how to a human this might end up being easier to check, just generating the finite loops programmatically its easier to do it the verbose way.

PR #222, opened 2026-07-15
Port csr_matmul_post.cpp to Kokkos

masterleinad:
`csr_matmul_post` was the next expensive kernel when running the `342` test case.


masterleinad:
Not clear if this is relevant for now.

PR #223, opened 2026-07-16
Remove "USE_INTEL_MKL" macro

quantumsteve:
In the last meeting we decided to ignore the batched gemm function in the Intel MKL.


PDoakORNL (approved):
LGTM

PR #228, opened 2026-07-21
Pr/gbek reference

PDoakORNL:
This PR adds python code that serves as a
Standalone reference implementation (numpy/scipy only, no dependency on
cincuenta, Ainur, or LanczosPlusPlus) that computes the "exact"
-i*Λ^+_<(t,t') hybridization for the atomic-limit hopping quench setup of

    Gramsch, Balzer, Eckstein, Kollar, PRB 88, 235106 (2013), Sec. VI.

This is the quantity plotted in the top-left panel of the paper's Fig. 3: a
self-consistently converged Weiss field, obtained by exact diagonalization of
a small SIAM, used as ground truth against which, in a future PR, cincuenta's production
rank-L Cholesky second-bath approximation (`NeqBathDecomposition.h`,
`ImpuritySolverNeqGBEK.h`, elsewhere in the `cincuenta` C++ tree) is
validated.

Python versions of the lowrank Cholesky and Eigenvector decomposition of the Weiss field
are also part of this PR and serve as reference/independent implementations of that portion 
of the methology as well.

This code is considerably less flexible than the cincuenta implementation and is limited to
quenches starting at the atomic limit for hopping and quenching with a non zero hopping at constant U.

More information is avaliable in the README.md

But generating the full report and all plots discussed in the documentation will not be possible until the cincuenta portion of the GBEK work is also merged.  Disentangling all the plots and their presentation in the report.tex and commiting just the part requiring the outputs of the python scripts from those that require cincuenta input seems like much work for little payout.

So as of now expect of the driving scripts only run_self_tests.sh to produce meaningful output.


g1257 (approved):
Approved, given that it doesn't run anything by default or pulls dependencies, so it basically stores Python scripts in the repo.

We may need to write a policy for Python scripts in the future, for cases they run by default. For now, I just note that this should not be a precedent for allowing scripts (independent of programming language) that pull things from the internet. I know we agree on this point, but stating it for the general audience.

PR #229, opened 2026-07-21
Port csr_kron_mult::imethod3

masterleinad:
This pull request ports csr_kron_mult::imethod3 to Kokkos by using an MDRangePolicy implementation.

I tried a bunch of more elaborate implementations but it seems we are mostly just bound by launch latencyand exposing as much parallelism as possible is advantageous. For the 342 benchmark, I'm seeing on Frontier
```
4.88e+00 sec 6.8% 100.0% 0.0% ------ 205199 csr_kron_mult::imethod3_pairs [for]
```
vs
```
9.37e+00 sec 22.6% 0.0% 0.0% 0.0% 0.00e+00 205199 PsimgLite::csr_kron_mult_method::imethod3 [region]
```
before. All the other kernels introduced here don't seem to play a big role even on CPUs where the profile looks like
```
BEGIN KOKKOS PROFILING REPORT:
TOTAL TIME: 23.771507 seconds
TOP-DOWN TIME TREE:
<average time> <percent of total time> <percent time in Kokkos> <percent MPI imbalance> <remainder> <kernels per second> <number of calls> <name> [type]
===================
|-> 1.25e+01 sec 52.7% 86.5% 0.0% 0.4% 3.30e+04 220232 PsimagLite::csr_kron_mult_method [region]
|   |-> 1.18e+01 sec 49.5% 92.0% 0.0% 5.5% 3.51e+04 206484 PsimgLite::csr_kron_mult_method::imethod3 [region]
|   |   |-> 1.08e+01 sec 45.3% 100.0% 0.0% ------ 206484 PsimgLite:csr_kron_mult::imethod3::spmv_kernel [for]
|   |   |-> 2.19e-01 sec 0.9% 0.0% 0.0% 100.0% 0.00e+00 206484 PsimgLite::csr_kron_mult_method::imethod3::copy_results [region]
|   |   |-> 7.29e-02 sec 0.3% 0.0% 0.0% 100.0% 0.00e+00 206484 PsimgLite::csr_kron_mult_method::imethod3::store_rows [region]
|   |   |-> 5.37e-02 sec 0.2% 100.0% 0.0% ------ 206484 Kokkos::View::initialization [PsimgLite:csr_kron_mult::imethod3::x_dev] via memset [for]
|   |-> 3.91e-01 sec 1.6% 0.0% 0.0% 100.0% 0.00e+00 8256 PsimgLite::csr_kron_mult_method::imethod2 [region]
|   |-> 3.09e-01 sec 1.3% 0.0% 0.0% 100.0% 0.00e+00 5492 PsimgLite::csr_kron_mult_method::imethod1 [region]
|-> 8.72e+00 sec 36.7% 99.4% 0.0% 0.4% 1.62e+04 70732 kokkos_gemm [region]
    |-> 8.67e+00 sec 36.5% 99.8% 0.0% 0.2% 7.23e+03 62673 KokkosBlas::gemm[ETI] [region]
    |   |-> 7.32e+00 sec 30.8% 100.0% 0.0% ------ 26647 KokkosBlas::gemm[NN] [for]
    |   |-> 8.01e-01 sec 3.4% 100.0% 0.0% ------ 23536 KokkosBlas::gemm[NT] [for]
    |   |-> 4.37e-01 sec 1.8% 100.0% 0.0% ------ 3996 KokkosBlas::gemm[CN] [for]
    |   |-> 9.64e-02 sec 0.4% 100.0% 0.0% ------ 8494 KokkosBlas::gemm[TN] [for]
```

### DIFF
--- a/dmrg/KronUtil/csr_kron_mult.cpp
+++ b/dmrg/KronUtil/csr_kron_mult.cpp
@@ -1,4 +1,7 @@
 #include "util.h"
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Profiling_ScopedRegion.hpp>
+#include <PsimagLite/KokkosType.h>
 
 #include <Kokkos_Profiling_ScopedRegion.hpp>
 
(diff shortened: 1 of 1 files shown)


PDoakORNL:
Unit test?  It seems like one can be produced for non trivial input, what corner cases are possible here?

PsimagLite/src/PsimagLite/KokkosType.h
```diff
@@ -0,0 +1,23 @@
...
+// The scalar types that are floating point types and their corresponding std::complex types.
+// We need to map std::complex to Kokkos::complex while keeping all the other types which is what
+// KokkosType does.
+template <typename T> struct KokkosType {
+	using type = T;
+};
+
+template <typename T> struct KokkosType<std::complex<T>> {
+	using type = Kokkos::complex<T>;
+};
```
quantumsteve, line 17: Is the repeated below? 
https://github.com/dmrgpp-project/dmrgpp/blob/5f1fee02104eada9f959aaae030e6f0176df0929/PsimagLite/src/PsimagLite/kokkos_gemm.cpp#L10-L25

masterleinad: Yes, I was planning to use the new header in `kokkos_gemm.cpp` in a follow-up but I don't mind much doing already here if y'all prefer.

quantumsteve:
> Unit test? It seems like one can be produced for non trivial input, what corner cases are possible here?

I'm also torn on this. It would take me some time to work through both kernels and that doesn't provide any future guarantee. We aren't resourced for software engineering perfection, but system testing alone aren't enough. What would a minimal unit test look like?

g1257:
@masterleinad : I'll let @quantumsteve work with you to address the unit test; perhaps it can be added given that it would only need standalone PsimagLite::CrsMatrix objects and PsimagLite::Matrix objects, it appears. FWIW, this PR works on my setup but I have no walltimes that are useful to compare.

masterleinad:
AFAICT, we already have unit tests, see https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/KronUtil/test1.cpp and https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/KronUtil/test2.cpp.

quantumsteve:
> AFAICT, we already have unit tests, see https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/KronUtil/test1.cpp and https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/KronUtil/test2.cpp.

aside: those tests will only fail on Debug builds.

quantumsteve:
@masterleinad Would you update these tests to fail if there are errors?
https://github.com/dmrgpp-project/dmrgpp/blob/0a2bd5ce8e51292fc091a6cffc259966bf89d421/dmrg/KronUtil/test1.cpp#L783-L893

PDoakORNL:
I don't want to muddy the waters further at this point, but this was my take on this yesterday:
https://github.com/masterleinad/dmrgpp/pull/1
I guess this wasn't a visible way to contribute.

My point in making that PR was basically writing testing shouldn't be hard and it shouldn't be something you push back on.

Also I don't think we should be continuing to work on unit tests that don't use Catch2. Probably someone should spend 100k tokens and port the existing tests to use Catch2.

masterleinad:
> @masterleinad Would you update these tests to fail if there are errors?

See #235.

> My point in making that PR was basically writing testing shouldn't be hard and it shouldn't be something you push back on.

I was under the impression that we had sufficient testing when working on this pull request.
As it turned out, that's not quite the case. Since this pull request is about performance improvements and doesn't add any new code paths, I don't think the testing situation should be improved here but rather in a separate pull request (#235). In the end, the existing test looks pretty good to me; it just didn't actually report test failures.

quantumsteve (approved):

g1257 (approved):

dmrg/KronUtil/csr_kron_mult.cpp
```diff
@@ -279,43 +282,121 @@ void csr_kron_mult_method(const int  imethod,
...
+			    KokkosScalar prod = cij * y_dev(iy, jy);
+			    Kokkos::atomic_add(&x_dev(ix, jx), prod);
+		    });
+		{
+			Kokkos::Profiling::ScopedRegion region(
+			    "PsimgLite::csr_kron_mult_method::imethod3::copy_results");
+			auto xhost
+			    = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace {}, x_dev);
+			for (int ix = 0; ix < nrow_X; ++ix) {
+				for (int jx = 0; jx < ncol_X; ++jx)
```
Rombur, line 395: You could use `parallel_for` with `MDRange` to use more threads if `OpenMP` is enable.

Rombur (approved):

PR #230, opened 2026-07-23
Remove USE_COMPLEX_Z

quantumsteve:
This removes the `USE_COMPLEX_Z` macro, building both `double` and `std::complex<double>` instantiations of each function. This also eliminates `MYTYPE` and `MAGMA_T` macros.

I also removed the `USE_MALLOC` macro and code contained as `dmrg_malloc` and `dmrg_free` do the same thing.

Lastly, I fixed some memory allocations preventing the batched plugin from working on multiple GPUs.


quantumsteve:
@g1257 What in dmrgpp decides if we are using the real or complex code? single or double precision?

g1257:
Complex is using only if either of the following conditions is met

(1) usecomplex is in SolverOptions in the input, or
(2) TimeStepTargeting is in SolverOptions

See https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/Engine/DmrgRunner.cpp#L70

If neither condition is met, then double is used.

quantumsteve:
> Complex is using only if either of the following conditions is met (1) usecomplex is in SolverOptions in the input, or (2) TimeStepTargeting is in SolverOptions See https://github.com/dmrgpp-project/dmrgpp/blob/master/dmrg/Engine/DmrgRunner.cpp#L70 If neither condition is met, then double is used.

OK, so instantiating both real and complex makes sense.

g1257:
Yes, instantiating both real and complex makes sense.

PDoakORNL:
This looks good to me but I would like to at least get a Magma CI environment ready and do a little testing on this.  Whether the new CI flavor gets commit before or after is probably less crucial.

dmrg/GPUPlugin/dmrg_magma.h
```diff
@@ -116,38 +98,38 @@ static void magmablas_Xgemm_vbatched_max_nocheck(magma_trans_t        transA,
...
 {
 
 	magmablas_zgemm_vbatched_max_nocheck(transA,
 	                                     transB,
 	                                     m,
 	                                     n,
 	                                     k,
-	                                     alpha,
-	                                     dA_array,
+	                                     *(magmaDoubleComplex*)&alpha,
```
masterleinad: I'd prefer to get rid of C-style pointer casts and rather use `reinterpret_cast` where necessary.
```suggestion
	                                     reinterpret_cast<magmaDoubleComplex&>(alpha),
```
here and elsewhere.

dmrg/GPUPlugin/setup_matrix.cpp
```diff
@@ -1,12 +1,18 @@
...
-#endif
+namespace {
+
+template <typename T> struct is_complex : std::false_type { };
+
+template <typename T> struct is_complex<std::complex<T>> : std::true_type { };
+
+template <typename T> inline constexpr bool is_complex_v = is_complex<T>::value;
+
+}
```
masterleinad: We already have `PsimagLite::IsComplexNumber` (defined in `PsimagLite/Complex.h`) so we don't need to reinvent it here.

masterleinad (approved):

PR #231, opened 2026-07-23
Fix PsimagLite::Matrix resize contract violation in LanczosPlusPlus::Combinatorial

PDoakORNL:
## Summary
- `Combinatorial::resize(m)` called `Matrix::resize(m,m,0)`, the 3-arg overload, which is NOT element-preserving across a change in row count (column-major flat-array truncation, no 2D remapping). Since `Combinatorial::comb_` is a process-lifetime static cache resized on every basis construction (`2*nsite+2`), constructing differently-sized bases across a process (e.g. different Catch2 test cases) could leave stale data in cells `doCombinatorial()` never explicitly writes (`comb_(n,m)` for `m>n`, mathematically 0), silently corrupting later `combinatorial()`/`perfectIndex()` queries.
- This caused a GBEK-solver test suite failure (in a separate in-flight PR) to curiously fall outside of numerical limits but only when running multiple test cases in a test program.
- Documented `Matrix::resize`'s two overloads' actual (quite different) contracts, and added `test_Matrix.cpp` locking both down.
- `LanczosPlusPlus` had no Catch2 test infrastructure at all before this (only legacy `TestSuite/inputs/`). Added the CMake wiring plus `test_Combinatorial.cpp`/`test_BasisOneSpin.cpp`, both verified to actually fail under the reverted (buggy) code before confirming they pass with the fix.

## Commentary
Statics + in place nondestructive resizing. These aren't good code smells. I'm aiming to just make the situation a bit less dangerous here and not exploring the underlying design issues.

🤖 Partially Generated and then complete rewritten since with [Claude Code](https://claude.com/claude-code)
mostly missed the point.


PsimagLite/src/PsimagLite/Matrix.h
```diff
@@ -266,13 +266,37 @@ template <typename T> class Matrix {
...
+	 * existing data to misinterpret). If you need old contents preserved (or
+	 * correctly zeroed) at their true (row, col) identity across a reshape
+	 * that changes nrow, use resize(nrow, ncol) instead.
+	 */
 	void resize(SizeType nrow, SizeType ncol, const T& val)
 	{
 		nrow_ = nrow;
 		ncol_ = ncol;
 		data_.resize(nrow * ncol, val);
 	}
```
g1257, line 288: Yes, thanks and sorry.
I likely forgot that std::vector::resize does not overwrite every location with val, but only the new ones and preserves the old ones. Very very bad on my part.

I may submit a PR that deletes this PsimagLite::Matrix::resize(3 args) entirely if it's not needed.

g1257 (approved):

PsimagLite/src/PsimagLite/Matrix.h
PDoakORNL: This was pretty subtle I would have not found it without the series of unit tests in a single test program and if I wasn't so suspicious of what Claude thought was wrong. It was convinced this was a heap corruption issue.

My quick scan of the code showed a few more uses of the 3 arg resize so I went with this pretty minimal fix.

PR #232, opened 2026-07-24
Deletes Matrix::resize with three arguments

g1257:
This PR deletes member function Matrix::resize(rows, cols, val), because it's dangerous to use; see this https://github.com/dmrgpp-project/dmrgpp/pull/231 for explanation; PR #231 will be merged before this one.

The other resize(rows, cols) with two arguments remains, because it is safely implemented.

### DIFF
--- a/dmrg/Engine/ObservableLibrary.h
+++ b/dmrg/Engine/ObservableLibrary.h
@@ -687,7 +687,7 @@ private:
 		}
 	}
 
-	void ppupupdndn(MatrixType& m,
+	void ppupupdndn(MatrixType& m1,
 	                MatrixType& m2,
 	                SizeType    orb1,
 	                SizeType    orb2,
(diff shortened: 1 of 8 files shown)


PDoakORNL:
There is just enough going on here that I want to read it fresh on Monday.

dmrg/Engine/ObservableLibrary.h
```diff
@@ -773,10 +773,8 @@ template <typename ObserverType> class ObservableLibrary {
 			}
 		}
 
-		m  = mup;
-		m2 = mdown;
-		// std::cout << mSinglet;
-		// std::cout << mTriplet;
+		m1 = std::move(mup);
```
PDoakORNL, line 776: Using this implicit move assignment operator is a new code path which  by my analysis should be fine.  I wouldn't normally ask for test coverage of an implicit method since my rule of thumb is objects have an obligation to either have working assign/move/copy or explicitly delete, protect private them to cause compilation time failure rather than bad behavior at runtime. 

That might be something we want to make clear in the dev docs.

PDoakORNL (changes requested):
I think 
dmrg/Engine/BlockOffDiagMatrix.h:40
Is a problem here, its just not apparent since its not actually instantiated.

Delete that constructor or perhaps add a unit test to protect this constructor and force its instantiation and therefore catch errors related to changes that impact it.

dmrg/Engine/ObservableLibrary.h
g1257: BlockOffDiagMatrix unused ctor is gone; good catch regarding that surviving resize(3) that wasn't being compiled in.

Regarding dmrg/Engine/ObservableLibrary.h line 776; I suppose all classes should have tests of their public interface. 

What we're seeing right now with DMRG++ is that many classes are too complicated to construct in a unit test, which limits unit tests that can be done with reasonable length. For the old and complicated classes the best we can do is full pipeline tests. 

Moving forward, classes should be simpler to construct so unit tests can be done.

PDoakORNL (approved):
This looks good to me now.

PR #234, opened 2026-07-24
Check text and numbers

g1257:
TextCheckAndNumbers will be used to test the KronLogger as a full pipeline test, and introspect, and other things that require full pipeline tests.

KronLogger appears too difficult for a unit test: it'd require an InitKron object, which requires in turn more non-trivial objects, a few levels down. It is the current situation for these old classes. Bottom line: we must compare two files: the one produced by the KronLogger with a reference file. TextCheckAndNumbers (this PR)  checks words, integers, and real numbers with the tolerance provided.

Creating a separate class for checking diffs between two files looks bad, but the advantage is we get exactly what's needed without dependencies. Let me know what you think, though.

### DIFF
--- a/PsimagLite/examples/checkTextAndNumbers.cpp
+++ b/PsimagLite/examples/checkTextAndNumbers.cpp
@@ -16,6 +16,16 @@ std::string file1;
 std::string file2;
 std::unique_ptr<PsimagLite::TextAndNumbersChecker> checker;
 
+std::string parseIgnoredLinePrefix(const char* text)
+{
+	const std::string argument(text);
+	const std::string option = "--ignore-line-prefix=";
+	if (argument.compare(0, option.size(), option) != 0 || argument.size() == option.size())
+		throw std::invalid_argument("Invalid option: \"" + argument + "\"");
+
+	return argument.substr(option.size());
+}
+
 double parseTolerance(const char* text)
 {
 	const std::string token(text);
@@ -42,15 +52,19 @@ TEST_CASE("text and numbers in two files match", "[TextAndNumbersChecker]")
 
 int main(int argc, char* argv[])
 {
-	if (argc != 4) {
-		std::cerr << "Usage: " << argv[0] << " FILE1 FILE2 TOLERANCE\n";
+	if (argc != 4 && argc != 5) {
+		std::cerr << "Usage: " << argv[0]
+		          << " FILE1 FILE2 TOLERANCE [--ignore-line-prefix=PREFIX]\n";
 		return EXIT_FAILURE;
 	}
 
 	try {
 		file1 = argv[1];
 		file2 = argv[2];
-		checker = std::make_unique<PsimagLite::TextAndNumbersChecker>(parseTolerance(argv[3]));
+		const std::string ignoredLinePrefix
+		    = (argc == 5) ? parseIgnoredLinePrefix(argv[4]) : "";
+		checker = std::make_unique<PsimagLite::TextAndNumbersChecker>(
+		    parseTolerance(argv[3]), ignoredLinePrefix);
 	} catch (const std::exception& error) {
 		std::cerr << error.what() << '\n';
 		return EXIT_FAILURE;
--- a/PsimagLite/examples/checkTextAndNumbers.cpp
+++ b/PsimagLite/examples/checkTextAndNumbers.cpp
@@ -12,8 +12,8 @@
 
 namespace {
 
-std::string file1;
-std::string file2;
+std::string                                        file1;
+std::string                                        file2;
 std::unique_ptr<PsimagLite::TextAndNumbersChecker> checker;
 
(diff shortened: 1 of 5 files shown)


PsimagLite/src/tests/testCheckTextAndNumbers.cpp
```diff
@@ -0,0 +1,384 @@
...
+				throw std::runtime_error("Cannot create test file: "
+				                         + path.string());
+
+			stream.write(contents.data(),
+			             static_cast<std::streamsize>(contents.size()));
+			if (!stream)
+				throw std::runtime_error("Cannot write test file: "
+				                         + path.string());
+
+			return path.string();
```
quantumsteve: Thoughts on returning `std::filesystem::path`?

PsimagLite/src/tests/testCheckTextAndNumbers.cpp
```diff
@@ -0,0 +1,384 @@
...
+		}
+
+		std::string path(const std::string& name) const
```
quantumsteve: return `std::filesystem::path`?

PsimagLite/examples/checkTextAndNumbers.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <iostream>
+#include <memory>
+#include <stdexcept>
+#include <string>
+
+namespace {
+
+std::string                                        file1;
+std::string                                        file2;
+std::unique_ptr<PsimagLite::TextAndNumbersChecker> checker;
```
quantumsteve: Can we avoid global variables?

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
```diff
@@ -0,0 +1,422 @@
...
+
+	//---------------------------------------------------------------------------//
+	/*!
+	 * \brief Determine whether a character is ASCII whitespace
+	 *
+	 * \param[in] c Character to classify
+	 *
+	 * \returns True for an ASCII whitespace character
+	 */
+	static bool isAsciiWhitespace(char c)
```
quantumsteve: https://en.cppreference.com/cpp/string/byte/isspace

quantumsteve (changes requested):
Consider std::string_view in place of `const std::string &`. 
https://en.cppreference.com/cpp/string/basic_string_view

On Linux, creating a std::string_view from `std::filesystem::path` doesn't require a memory allocation.

```
std::filesystem::path p = "/usr/bin/g++";
auto native_view = std::string_view(p.native()); 
```

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
```diff
@@ -0,0 +1,400 @@
...
+	 * \param[out] token     Token read from the stream
+	 *
+	 * \returns True if a token was read, or false at end of stream
+	 */
+	static bool readToken(std::istream& stream, std::string& token)
+	{
+		token.clear();
+		char c = 0;
+		while (stream.get(c)) {
+			if (!std::isspace(c)) {
```
quantumsteve:
```suggestion
			if (!std::isspace(static_cast<unsigned char>(c))) {
```

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
```diff
@@ -0,0 +1,400 @@
...
+				token.push_back(c);
+				break;
+			}
+		}
+
+		if (token.empty())
+			return false;
+
+		while (stream.get(c)) {
+			if (std::isspace(c))
```
quantumsteve:
```suggestion
			if (std::isspace(static_cast<unsigned char>(c)))
```

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
```diff
@@ -0,0 +1,400 @@
...
+	/*!
+	 * \brief Compare two text files token by token
+	 *
+	 * \param[in] file1 Path to the first file
+	 * \param[in] file2 Path to the second file
+	 *
+	 * 	hrows std::runtime_error if a file cannot be opened, a numeric token
+	 *         cannot be represented, or the file contents do not match
+	 */
+	void run(const std::string& file1, const std::string& file2) const
```
quantumsteve:
```suggestion
	void run(const std::filesystem::path& file1, const std::filesystem::path& file2) const
```

PDoakORNL:
With the change to remove the global variables and catch dependence from the checkTextAndNumbers I don't think I have anything I need to see here past what @quantumsteve is already asked for.

quantumsteve (approved):

quantumsteve:
@g1257 missing header?

PR #235, opened 2026-07-27
Make test1.cpp use Catch2

masterleinad:
Wr noticed in https://github.com/dmrgpp-project/dmrgpp/pull/229#issuecomment-5074141356 that we do have tests for `csr_kron_mult` and `den_kron_mult` but it neither uses `Catch2`, nor errors out on failure or tests with `std::complex`. This pull request fixes that.
Sample output when making `imethod=3` a no-op:
```
Randomness seeded to: 2133787757

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
KronUtil_Test1 is a Catch2 v3.11.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
kron_mult_test1_double
-------------------------------------------------------------------------------
/tmp/dmrgpp/dmrg/KronUtil/test1.cpp:1198
...............................................................................

/tmp/dmrgpp/dmrg/KronUtil/test1.cpp:618: FAILED:
  REQUIRE( diffmax <= tol )
with expansion:
  0.20663306635186229 <= 0.000000001
with message:
  csr: transA=N, itransA 0 transB=N, itransB 0 nrow_A 1 ncol_A 1 nrow_B 1
  ncol_B 1
  ix 0, jx 0, diff1 0, diff2 0, diff3 0.206633

-------------------------------------------------------------------------------
kron_mult_test1_complex
-------------------------------------------------------------------------------
/tmp/dmrgpp/dmrg/KronUtil/test1.cpp:1200
...............................................................................

/tmp/dmrgpp/dmrg/KronUtil/test1.cpp:618: FAILED:
  REQUIRE( diffmax <= tol )
with expansion:
  0.3228655156206291 <= 0.000000001
with message:
  csr: transA=N, itransA 0 transB=N, itransB 0 nrow_A 1 ncol_A 1 nrow_B 1
  ncol_B 1
  ix 0, jx 0, diff1 0, diff2 0, diff3 0.322866

===============================================================================
test cases:  2 |  0 passed | 2 failed
assertions: 18 | 16 passed | 2 failed
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,6 +6,7 @@ UseTab: ForIndentation
 TabWidth: 8
 IndentWidth: 8
 BreakBeforeBraces: Custom
+BreakStringLiterals: false
 IndentAccessModifiers: false
 AccessModifierOffset: -8
 EmptyLineAfterAccessModifier: Always
--- a/dmrg/KronUtil/test1.cpp
+++ b/dmrg/KronUtil/test1.cpp
@@ -401,64 +401,34 @@ template <typename T> void run_kron_checks()
 											if (diffmax
 											    > tol) {
 												INFO(
-												    "den: "
-												    "transA"
-												    "="
+												    "den: transA="
 												    << transA
-												    << ", "
-												       "itr"
-												       "ans"
-												       "A "
+												    << ", itransA "
 												    << itransA
-												    << " tr"
-												       "ans"
-												       "B="
+												    << " transB="
 												    << transB
-												    << ", "
-												       "itr"
-												       "ans"
-												       "B "
+												    << ", itransB "
 												    << itransB
-												    << " nr"
-												       "ow_"
-												       "A "
+												    << " nrow_A "
 												    << nrow_A
-												    << " nc"
-												       "ol_"
-												       "A "
+												    << " ncol_A "
 												    << ncol_A
-												    << " nr"
-												       "ow_"
-												       "B "
+												    << " nrow_B "
 												    << nrow_B
-												    << " nco"
-												       "l_"
-												       "B "
+												    << " ncol_B "
 												    << ncol_B
 												    << '\n'
 												    << "ix "
 												    << ix
-												    << ", "
-												       "jx "
+												    << ", jx "
 												    << jx
-												    << ", "
-												       "dif"
-												       "f12"
-												       " "
+												    << ", diff12 "
 												    << diff12
-												    << ", "
-												       "dif"
-												       "f23"
-												       " "
+												    << ", diff23 "
 												    << diff23
-												    << ", "
-												       "dif"
-												       "f31"
-												       " "
+												    << ", diff31 "
 												    << diff31
-												    << " di"
-												       "ff4"
-												       "1 "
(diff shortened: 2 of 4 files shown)


dmrg/KronUtil/test1.cpp
```diff
@@ -1017,69 +1097,91 @@ int main()
...
-												printf(
-												    "ix %d, jx %d, diff1 %f, diff2 %f, diff3 %f \n",
-												    ix,
-												    jx,
-												    diff1,
-												    diff2,
-												    diff3);
+											if (diffmax
+											    > tol) {
+												INFO(
```
PDoakORNL, line 1022: What is going on with the formatting or output  here? This is an insane way to format text.

PDoakORNL (commented):
Ok so this test modernization is respectful of the continuity of the project but I just want to make sure I understand the failure that you're pointing out, this isn't easy to read through at all do to the formatting.  Is it just a problem with the tests that the std::complex variant fails or is it a problem with the underlying implementation?

masterleinad:
> Is it just a problem with the tests that the std::complex variant fails or is it a problem with the underlying implementation?

It doesn't fail. I just wanted to point out what the output looks like if modifying one of the `crs_kron_mult` implementations to do nothing.

dmrg/KronUtil/test1.cpp
masterleinad: That's just what `clang-format` wanted with the current settings. We can of course just disable it locally.

dmrg/KronUtil/test1.cpp
```diff
@@ -2,33 +2,28 @@
...
-	int            ncol_B            = 0;
-	int            itransA           = 0;
-	int            itransB           = 0;
-
-	static const bool           needsPrinting   = false;
-	const SizeType              gemmRnb         = 49;
-	const SizeType              threadsForGemmR = 1;
-	PsimagLite::GemmR<RealType> gemmR(needsPrinting, gemmRnb, threadsForGemmR);
+	using RealT                   = typename PsimagLite::Real<T>::Type;
+	const RealT denseFlopDiscount = RealT(0.2);
```
quantumsteve:
```suggestion
	const RealT denseFlopDiscount = 0.2;
```

Do we have a preference on initialization? I don't think the extra cast here provides any benefit. Brace initialization adds a narrowing check but looks less like math.

.clang-format
```diff
@@ -6,6 +6,7 @@ UseTab: ForIndentation
 TabWidth: 8
 IndentWidth: 8
 BreakBeforeBraces: Custom
+BreakStringLiterals: false
```
PDoakORNL, line 9: I'm happy to have this added to the PR even though its a bit of a scope drift. I didn't notice until this PR how obnoxious `BreakStringLiterals: true` was in this code base.

PDoakORNL (approved):
This looks good enough for me now.

quantumsteve (approved):

PR #236, opened 2026-07-27
Make warning set for known compilers part of CMake

PDoakORNL:
## Summary
- Adds a `DMRG_WARNINGS_AS_ERRORS` CMake option (default `ON`) that mirrors CI's warning settings (`.github/workflows/ci.yml`, `ci-mac.yml`) locally: `-Wall -Wextra -Wno-unused-parameter` for GNU/Clang, plus `CMAKE_COMPILE_WARNING_AS_ERROR` when the option is on.
- Previously these flags only existed in the CI workflow files, not in the project's own CMakeLists.txt — a local build would need to pass `-DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-unused-parameter" -DCMAKE_COMPILE_WARNING_AS_ERROR=ON` explicitly to see what CI would flag.
- Removes the now-redundant flag lines from `ci.yml`/`ci-mac.yml`, since the default now lives in one place (the root `CMakeLists.txt`).
- Pass `-DDMRG_WARNINGS_AS_ERRORS=OFF` to opt out.

## Discussion
Defaulting this ON changes the out-of-the-box build behavior for every contributor — turning compiler warnings into hard build failures locally, matching what already happens in CI. Open to feedback on whether ON or OFF should be the default here.


CMakeLists.txt
```diff
@@ -59,6 +59,17 @@ else()
   set(CMAKE_INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}")
 endif()
 
+# Mirror CI's warning settings (.github/workflows/ci.yml, ci-mac.yml) so
```
masterleinad: That's not true anymore after this pull request. 🙂

masterleinad (commented):
I'm fine with always enabling warnings we acre about but wouldn't add an option that does the exact same thing as (`CMAKE_COMPILE_WARNING_AS_ERROR`).

CMakeLists.txt
PDoakORNL: Thats true, I changed my mind part way through this PR.  If we get consensus on this I'll rewrite the documentation here to make sense.

PDoakORNL:
After reading the discussion around the [merge of CMAKE_COMPILE_WARNING_AS_ERROR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7187) I'm having second thoughts about any method of defaulting compile warnings as error on.  I may cut this down to just the matching warnings setup.

quantumsteve:
We don't want users to be treating compile warnings as errors.

dalg24 (commented):
FWIW I don't like adding an option for this, I think it belongs to user configuration land

PDoakORNL:
Ok so not every Friday night idea is great.  I will just remember to set CMAKE_COMPILE_WARNING_AS_ERROR for my builds.  I do still think there is some value in capturing the warning set we are using in the CMake and it might improve our chances of getting issues filed on or noticing warnings with new/different versions of clang/gcc.  This won't cause breakage on compilers using different options for warnings.

masterleinad (approved):
Fine with me.

quantumsteve (approved):

PR #237, opened 2026-07-27
Follow naming convention for Kokkos-related files

masterleinad:
Following https://github.com/dmrgpp-project/dmrgpp/pull/229/changes#r3633706140, this pull request also adds `PsimagLite/KokkosType.h`.


PsimagLite/src/PsimagLite/KokkosGemm.cpp
```diff
@@ -5,24 +5,8 @@
...
-
-template <typename T>
-        requires(!std::is_floating_point_v<T>)
-struct KokkosType<T> {
-	using type = Kokkos::complex<typename T::value_type>;
-};
-
-}
+#include <PsimagLite/KokkosGemm.h>
+#include <PsimagLite/KokkosType.h>
```
quantumsteve:
```suggestion
#include "PsimagLite/KokkosGemm.h"
#include "PsimagLite/KokkosType.h"
```

Are we using quotes for headers in the same directory?

masterleinad: I don't see a reason for that. These files are visible under their respective include path.

PDoakORNL: My understanding was we said we were.  Part of me really wants there to be an actual point to <> and "" both being valid include syntax.

The idea is that it provides the reader with information about whether a header is external or not.  That said I don't actually look at includes when I read code unless something is broken, and the tooling I use doesn't care about this.

@g1257 ?

g1257:
On 7/27/26 3:17 PM, Peter Doak wrote:
> @PDoakORNL commented on this pull request.
> 
> 
> 
>> +#include <PsimagLite/KokkosGemm.h>
> +#include <PsimagLite/KokkosType.h>
> 
> My understanding was we said we were.  @g1257
> 

The file in questions seems to be
```PsimagLite/src/PsimagLite/KokkosGemm.cpp```
so the directory is
```PsimagLite/src/PsimagLite/```
but the file
```PsimagLite/src/PsimagLite/PsimagLite/KokkosGemm.h```
doesn't exist, so double quotes can't be used, and angle brackets must be used.

Another alternative would be
```
#include "KokkosGemm.h"
```
etc.

Either way is fine, but not 
```
#include "PsimagLite/KokkosGemm.h"
```
at least not for this particular file.
Did I get it right?

PsimagLite/src/PsimagLite/KokkosGemm.cpp
quantumsteve: For reference: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rs-incform

I may have been wrong to include the PsimagLite prefix.

g1257:
Oh, I see, without the prefix and with quotes is better as it's an internal reference (PsimagLite include from PsimagLite), you are correct.
We're supposed to use then
#include "KokkosGemm.h"
etc.

quantumsteve:
> Oh, I see, without the prefix and with quotes is better as it's an internal reference (PsimagLite include from PsimagLite), you are correct. We're supposed to use then #include "KokkosGemm.h" etc.

Yes, that how I'm reading the C++ core guidelines. Any of the above options seem to build without issue, so it's just about keeping a consistent style.

PsimagLite/src/PsimagLite/KokkosGemm.cpp
masterleinad: I'm used to provide full include paths for everything so I always know where to look for the file and never need to consider local behavior. Either way, I changed it.

dalg24 (approved):

quantumsteve (approved):

PR #238, opened 2026-07-27
Cincuenta license equals DMRG++'s

### LINKED ISSUE #211: cincuenta GPL License
Is this a problem?
It's not a problem for me, but cincuenta isn't on DOE Code and this makes we question whether we can just surface it as 
part of dmrgpp.  

```
cincuenta is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
cincuenta is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
GNU General Public License for more details.
You should have received a copy of the GNU General Public License
along with cincuenta. If not, see <https://www.gnu.org/licenses/>.
The full software license for cincuenta version 1.0.0
can be found in
file LICENSE.
```

Ways forward although obviously I'm no lawyer.
If everyone who's contributed to cincuenta agrees to switch the license we could just have it assume the DMRG license.

g1257:
Resolves https://github.com/dmrgpp-project/dmrgpp/issues/211


PDoakORNL (approved):
Looks good

PR #239, opened 2026-07-28
Refactor test2.cpp to use Catch2

masterleinad:
Similar to https://github.com/dmrgpp-project/dmrgpp/pull/235, test2.cpp wasn't using `Catch2` and didn't signal test failures. This pull request fixes that.

### DIFF
--- a/dmrg/KronUtil/test2.cpp
+++ b/dmrg/KronUtil/test2.cpp
@@ -1,31 +1,22 @@
 #include "KronUtil.h"
 #include "util.h"
 
-#ifndef USE_FLOAT
-using RealType = double;
-#else
-using RealType = float;
-#endif
+#include <Kokkos_Core.hpp>
+#define CATCH_CONFIG_RUNNER
+#include <catch2/catch_session.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <complex>
 
-int main()
+template <typename ComplexOrRealType> void run_kron_submatrix_checks()
 {
-	using RealType          = RealType;
-	using ComplexOrRealType = std::complex<RealType>;
-	int      nerrors        = 0;
-	RealType thresholdA     = 0;
-	RealType thresholdB     = 0;
-	int      nrow_A         = 0;
-	int      ncol_A         = 0;
-	int      nrow_B         = 0;
-	int      ncol_B         = 0;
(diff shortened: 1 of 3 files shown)


dalg24 (commented):
Merge both test 1 and 2 into a single executable?

masterleinad:
> Merge both test 1 and 2 into a single executable?

We can look into that in a follow-up.

dmrg/KronUtil/test2.cpp
```diff
@@ -293,43 +246,35 @@ int main()
...
+										     << " ncindex "
+										     << ncindex
+										     << '\n'
+										     << " ie " << ie
+										     << " je " << je
+										     << " eij "
+										     << eij
+										     << " dij "
+										     << dij);
+										REQUIRE(eij == dij);
```
quantumsteve, line 279: Should this be outside the `if` statement?

masterleinad: No, that's intended. We want to print what we store in `INFO` if the check fails but only construct the string if the check actually fails. We are using the same pattern in test1.cpp now.

dmrg/KronUtil/test2.cpp
```diff
@@ -1,31 +1,22 @@
...
-		for (thresholdA = 0; thresholdA <= 1.1; thresholdA += 0.1) {
-			for (ncol_A = 1; ncol_A <= 10; ncol_A += 3) {
-				for (nrow_A = 1; nrow_A <= 10; nrow_A += 3) {
-					for (ncol_B = 1; ncol_B <= 10; ncol_B += 3) {
-						for (nrow_B = 1; nrow_B <= 10; nrow_B += 3) {
-
+	using RealType = typename PsimagLite::Real<ComplexOrRealType>::Type;
+
+	for (RealType thresholdB = 0; thresholdB <= 1.1; thresholdB += 0.1) {
+		for (RealType thresholdA = 0; thresholdA <= 1.1; thresholdA += 0.1) {
```
quantumsteve: I think best practice is to use an integer loop variable and calculate the floating point value?

masterleinad: Sure. It feels a little out-of-scope, though.

quantumsteve (approved):

PR #241, opened 2026-07-31
Follow through on MPI in the dmrgpp Repo

PDoakORNL:
This adds MPI support for real to the cmake, adds actual MPI smoke testing for manyOmegas and adds a MPI ci flavor.  This is needed because I need to really start using MPI in cincuenta and support for that will follow.


dmrg/CMakeLists.txt
```diff
@@ -70,6 +70,47 @@ install(TARGETS dmrg observe toolboxdmrg omegas introspect RUNTIME DESTINATION $
...
+  add_mpi_test(omegas_mpi_dry_run 2 $<TARGET_FILE:omegas> -d -f ${PATH_TO_INPUTS}/inputManyOmegasDryRun.ain)
+  set_tests_properties(omegas_mpi_dry_run PROPERTIES PASS_REGULAR_EXPRESSION "MPI rank=1")
+
+  set(_omegas_mpi_tiny_workdir ${CMAKE_CURRENT_BINARY_DIR}/omegas_mpi_tiny_real)
+  file(MAKE_DIRECTORY ${_omegas_mpi_tiny_workdir})
+
+  add_test(NAME omegas_mpi_tiny_clean COMMAND ${CMAKE_COMMAND} -E rm -f tinyGs.hd5 tinyOmega0.hd5 tinyOmega1.hd5)
+  set_tests_properties(omegas_mpi_tiny_clean PROPERTIES LABELS "MPI" WORKING_DIRECTORY ${_omegas_mpi_tiny_workdir})
+
+  add_test(NAME omegas_mpi_tiny_gs COMMAND $<TARGET_FILE:dmrg> -f ${PATH_TO_INPUTS}/inputManyOmegasTinyGs.ain)
```
masterleinad, line 93: Do we need to check the output in any way?

dmrg/CMakeLists.txt
```diff
@@ -70,6 +70,47 @@ install(TARGETS dmrg observe toolboxdmrg omegas introspect RUNTIME DESTINATION $
...
+    $<TARGET_FILE:omegas>
+    -f
+    ${PATH_TO_INPUTS}/inputManyOmegasTiny.ain
+    -O
+    tinyOmega)
+  set_tests_properties(omegas_mpi_tiny_real PROPERTIES WORKING_DIRECTORY ${_omegas_mpi_tiny_workdir} DEPENDS
+                                                       omegas_mpi_tiny_gs PASS_REGULAR_EXPRESSION "MPI rank=1")
+
+  add_test(NAME omegas_mpi_tiny_outputs
+           COMMAND sh -c "test -f tinyGs.hd5 && test -f tinyOmega0.hd5 && test -f tinyOmega1.hd5")
```
masterleinad, line 109: If these output files are from multiple tests, there should also be multiple checks, right?

PsimagLite/src/CMakeLists.txt
```diff
@@ -47,20 +47,18 @@ target_link_libraries(psimaglite PUBLIC loki)
 
 add_library(psimaglite::psimaglite ALIAS psimaglite)
 
-# find_package(MPI COMPONENTS CXX)
-if(MPI_FOUND)
+if(USE_MPI)
   target_compile_definitions(psimaglite PUBLIC USE_MPI)
```
masterleinad, line 51: That would be a candidate for a configuration header.

CMakeLists.txt
```diff
@@ -60,6 +60,10 @@ else()
 endif()
 
 option(DMRG_BUILD_BATCHED "Enable building with Batched support" ON)
+option(USE_MPI "Enable MPI support" OFF)
```
masterleinad, line 63:
```suggestion
option(DMRGPP_USE_MPI "Enable MPI support" OFF)
```
to avoid clashes with any other (future) packages using the build system.

g1257 (approved):

dmrg/CMakeLists.txt
PDoakORNL: We do in a dependent check test.  But I have cleaned this all up a bit so that should be clear now.

dmrg/CMakeLists.txt
PDoakORNL: yes and should be clear now we're doing that.

PsimagLite/src/CMakeLists.txt
PDoakORNL: I'd rather take the path that this define be "namespaced" as you suggest below and that targets using the headers must also express that dependency in cmake and therefore will get the define.

CMakeLists.txt
PDoakORNL: good idea done

.github/workflows/ci-mpi.yml
```diff
@@ -0,0 +1,51 @@
+name: CI - Ubuntu 24.04 MPI
+
+on:
+  push:
+    branches:
```
masterleinad, line 5: Not sure if we really always want to run for all branches but we do that elsewhere as well. Thus. we should discuss that somewhere else.

.github/workflows/ci-mpi.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Configure
+      shell: spack-bash {0}
+      run: |
+        spack env activate ${{github.workspace}}/source
+        cmake -S ${{github.workspace}}/source \
+              -B ${{github.workspace}}/build \
+              -GNinja \
+              -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+              -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-unused-parameter" \
```
masterleinad:
```suggestion
```
That's the default now.

masterleinad (approved):

.github/workflows/ci-mpi.yml
PDoakORNL: I just copied that but I agree we probably don't.

.github/workflows/ci-mpi.yml
PDoakORNL: You're correct we don't want this CI flavor to be special wrt warnings.

PR #243, opened 2026-08-03
Add cuda ci flavors

PDoakORNL:
Continuing CI updates, testing github's reaction to this workflow.


PDoakORNL:
This is stacked on #241 and I will revise a bit further after that PR is merged.

masterleinad:
> This is stacked on #241 and I will revise a bit further after that PR is merged.

Isn't adding a MPI CI build orthogonal?

.github/workflows/ci-cuda.yml
```diff
@@ -0,0 +1,58 @@
+name: CI - CUDA A30
+
+on:
+  workflow_dispatch:
```
masterleinad, line 4: Why do you have this trigger?
I'm used to run CI on push to `master`.

.github/workflows/ci-cuda.yml
```diff
@@ -0,0 +1,58 @@
...
+  approve-external-pr:
+    name: Approve external PR for CUDA runner
+    if: >-
+      github.event_name == 'pull_request' &&
+      !contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
+    runs-on: ubuntu-24.04
+    environment: cuda-ci-approval
+    steps:
+    - name: Approval received
+      run: echo "External PR approved for CUDA CI"
```
masterleinad, line 23: How does this `job` work? Do have any workflow run where it's not skipped?

.github/workflows/ci-cuda.yml
```diff
@@ -0,0 +1,58 @@
...
+  cancel-in-progress: false
+
+jobs:
+  approve-external-pr:
+    name: Approve external PR for CUDA runner
+    if: >-
+      github.event_name == 'pull_request' &&
+      !contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
+    runs-on: ubuntu-24.04
+    environment: cuda-ci-approval
```
masterleinad, line 20: Where is this `environment` defined?

.github/workflows/ci-cuda.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+    - name: Approval received
+      run: echo "External PR approved for CUDA CI"
+
+  build:
+    name: RelWithDebInfo (CUDA, A30)
+    needs: approve-external-pr
+    if: >-
+      always() &&
```
masterleinad, line 29: Do you mean to use `always()` here? Can you explain why?

scripts/ci/run_cuda_step.sh
```diff
@@ -0,0 +1,101 @@
...
+  exit 2
+fi
+
+# This workflow runs on a self-hosted GPU VM.  Dependencies are expected to be
+# provided by a prebuilt, host-local Spack environment rather than rebuilt for
+# each CI run.  Override these paths with GitHub Actions variables or runner
+# environment variables if the VM layout changes.
+spack_setup="${DMRGPP_CUDA_SPACK_SETUP:-/scratch/spack/share/spack/setup-env.sh}"
+spack_env="${DMRGPP_CUDA_SPACK_ENV:-/scratch/spack_envs/dmrgpp}"
+cuda_root="${DMRGPP_CUDA_ROOT:-/usr/local/cuda-12.6}"
```
masterleinad, line 18: Are you intending to use this script outside of the CI? For the CI, it seems we are stroing all these settings already in the repository.

.github/workflows/ci-cuda.yml
PDoakORNL: This is not a needed feature, I'll drop it.

.github/workflows/ci-cuda.yml
PDoakORNL: Yes, this job runs for PR authors outside those three trusted GitHub associations. It is the
 approval gate before allowing their code onto the self-hosted CUDA runner. I’ll simplify the expression to make that clearer.

.github/workflows/ci-cuda.yml
PDoakORNL: This is setup for the project in the github UI.  This part was pretty clunky but it should let us use the github.com UI instead of using a hacky method like waiting on a comment from a trusted user that matches a particular regex.

.github/workflows/ci-cuda.yml
PDoakORNL: Just for the record I am not this knowledgeable about github runners, this would remain unknown to me without GPT-5.6.

For a trusted PR, the approval job is intentionally skipped. GitHub Actions normally skips a job when one of its "needs: jobs" was skipped, before evaluating the rest of its condition. always() tells Actions to evaluate the build condition anyway.

 Then the remainder says:

 - trusted author → run CUDA build;
 - untrusted author → run only if the approval job succeeded.

So always() does not bypass the approval gate; it only prevents trusted PRs from being skipped because their unnecessary
 approval job was skipped.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#example-not-requiring-successful-dependent-jobs

scripts/ci/run_cuda_step.sh
PDoakORNL: So here the idea is that we have fallbacks that match the CI machine configuration without having to go into the github GUI or API and set these variables.  So you could for instance in a fork setup your own CI and set these variables DMRGPP_CUDA_ROOT, etc. for the runner and your own CI runner.  This sounds unnecessary as I write it out.

scripts/ci/run_cuda_step.sh
```diff
@@ -0,0 +1,101 @@
...
+    if [[ -n "${CUDA_HOME:-}" ]]; then
+      cuda_root_args+=("-DCUDAToolkit_ROOT=${CUDA_HOME}")
+    fi
+
+    cmake -S "${source_dir}" \
+          -B "${build_dir}" \
+          -GNinja \
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DCMAKE_CXX_COMPILER=/usr/bin/g++-13 \
+          -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-unused-parameter" \
```
masterleinad:
```suggestion
```

scripts/ci/run_cuda_step.sh
```diff
@@ -0,0 +1,101 @@
...
+          -DCMAKE_COMPILE_WARNING_AS_ERROR=ON \
+          -DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=NEVER \
+          -DKokkos_ENABLE_CUDA=ON \
+          -DKokkos_ARCH_AMPERE80=ON \
+          -DKokkos_ENABLE_CUDA_LAMBDA=ON \
```
masterleinad: Not necessary and might result in errors for later Kokkos versions.
```suggestion
```

scripts/ci/run_cuda_step.sh
PDoakORNL: But I'd like to leave these overrides intact for forks or alternate cuda-versions

masterleinad (approved):

quantumsteve (approved):

quantumsteve:
Any idea why `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON` is set yet there are many warning messages in the logs?

masterleinad:
> Any idea why `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON` is set yet there are many warning messages in the logs?

`CMAKE_COMPILE_WARNING_AS_ERROR` just sets the C++ flag "-Werror" but not the CUDA one `-Werror=all-warnings`.

masterleinad (approved):

PR #245, opened 2026-08-05
Always build with batched support

masterleinad:
I don't see a good reason to make this an option.
Note that this is somewhat of a misnomer: https://github.com/dmrgpp-project/dmrgpp/blob/9438824a58481b6d1a69add55de96c72799b3909/dmrg/Engine/MatrixVectorKron/BatchedGemmInclude.hh#L1-L11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,10 +65,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   add_compile_options(-Wall -Wextra -Wno-unused-parameter)
 endif()
 
-option(DMRG_BUILD_BATCHED "Enable building with Batched support" ON)
-
-include(CMakeDependentOption)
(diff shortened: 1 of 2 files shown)


g1257:
Yes, this is fine. This code path will always be compiled in but only activated if BatchedGemm is in SolverOptions then; which is the default even now. 
I want to find time to work a bit on decreasing compile times (worst case for the release one is around 6 minutes right now); see also https://github.com/dmrgpp-project/dmrgpp/issues/246

masterleinad:
Given what the option actually does, i.e., selecting the plugin path for the `BatchedGemm` implementation, we should probably rather just rename the option. Even with the option turned off, the `BatchedGemm` solver setting is still supported.

quantumsteve (approved):
`DMRG_BUILD_BATCHED` is an artifact from GPUPlugin being separate.

masterleinad:
Let's go with https://github.com/dmrgpp-project/dmrgpp/pull/247 instead.

PR #253, opened 2026-08-07
Eliminate unused alpha, beta, transa and transb arrays

quantumsteve:
While replacing dmrg_malloc with Kokkos::view and kokkos_malloc, it looks like several arrays are unused.


dmrg/GPUPlugin/dmrg_magma.h
```diff
@@ -139,6 +139,47 @@ static void magmablas_Xgemm_vbatched_max_nocheck(magma_trans_t
...
+	    *reinterpret_cast<magmaDoubleComplex*>(&beta),
+	    reinterpret_cast<magmaDoubleComplex**>(dC_array),
+	    lddc,
+	    batchCount,
+	    max_m,
+	    max_n,
+	    max_k,
+	    queue);
+}
+
```
masterleinad, line 141: Do we need both the interface using `std::complex` and the one using `Kokkos::complex`? Is the former one still used after this pull request?

masterleinad (approved):

PR #259, opened 2026-08-13
Update README

g1257:


HowToCiteDMRGPP.md
```diff
@@ -0,0 +1,50 @@
+## How To Cite This Work
```
PDoakORNL, line 1: Apparently the "github" way to do this is to use a 
CITATION.cff file.
(https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files)
They can look quite similar to this although I guess are technically yaml but I think we probably should do that since the machine lords seem to be standardizing on that and its little different from this.

PDoakORNL (approved):
Consider the HowToCiteDMRGPP.md->CITATION.cff and the small formatting change around the bibtex

PR #261, opened 2026-08-13
Document and clean up the non-equilibrium DMFT ED input

PDoakORNL:
## Summary

- Documents the non-equilibrium DMFT (neq-DMFT) exact-diagonalization solver's input format for a new, non-developer user, and completes the previously incomplete equilibrium input reference table.
- Locates cincuenta's design docs/diagrams and input documentation in cincuenta/doc
- Cleans up `NeqSolver=` input handling: collapses the previous three-way `tdmrg`/`gbek`/bare-ED dispatch down to two real choices, `"ed"` (default) and `"tdmrg"`. `ImpuritySolverNeqGBEK` at `NeqBathRank=0` provably delegates entirely to `ImpuritySolverNeqExactDiag`, so bare ED was never a distinct method -- just the L=0 case of GBEK exposed as its own solver string. Unrecognized `NeqSolver=` values now error instead of silently falling through to ED.
- Fixes a real inconsistency: `NeqAtomicLimit=1` forced the lattice-side hopping to zero unconditionally, but the impurity-side bath substitution only happened on the old `"gbek"` branch -- so a bare-ED run with `NeqAtomicLimit=1` combined a zero-hopping lattice with a nonzero-bath impurity Hamiltonian. `NeqSolver="tdmrg"` + `NeqAtomicLimit=1` (untested) is now rejected with a clear error instead.
- Skips the equilibrium bath fit entirely when `NeqAtomicLimit=1`, since its result was always discarded -- removes the need for a placeholder `NumberOfBathPoints>=1` just to keep the bath-fit minimizer from erroring on zero free parameters.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PDoakORNL:
On further thinking about this I tightened up the consistency of the neq input further and no longer accept nonsense input for `ChemicalPotential` in the `NeqAtomicLimit=1` case or when `FittingOptions=ParticleHoleSymmetric`. Previously you could just put whatever value you wanted in there and the code used U/2 for ChemicalPotential on its own.  I personally really dislike this sort of behavior from Scientifc codes. If it is in the input it should do something and it should not contradict other input.

g1257 (approved):

PR #263, opened 2026-08-14
Build script for end users

### LINKED ISSUE #260: Create optional compilation script
For end-users that want a managed solution after they have cloned until the executables are produced.

One script to rule them all:
```
./compile
```
[create build directory; asks whether user wants cpu_only; if yes, determines Linux or mac and runs cmake with correct options; else asks user whether AMD or NVIDIA and runs cmake with correct options; may run make and install also automatically]
 
Supports unattended run
```
./compile --non-interactive  --gpu=amd
```

g1257:
DMRG++ is distributed in source code form. This PR creates a build script that helps end users after the git clone (after they have the source) until and including building the executables: it creates a build directory, runs cmake with correct options for release;  GPU support will be added in the future.

Closes https://github.com/dmrgpp-project/dmrgpp/issues/260


configure.sh
```diff
@@ -0,0 +1,41 @@
...
+source_dir=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+build_dir="${source_dir}/build"
+
+mkdir -p "${build_dir}"
+
+cmake_options=(
+    -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_COMPILE_WARNING_AS_ERROR=ON
+    -DKokkosKernels_ENABLE_TPL_BLAS=ON
+    -DKokkosKernels_ENABLE_TPL_LAPACK=ON
```
masterleinad: You don't need that anymore.
```suggestion
```

configure.sh
```diff
@@ -0,0 +1,41 @@
...
+)
```
masterleinad: Do end users need any of that? Isn't it enough to just call `cmake` without any extra arguments? `Release` vs `RelWithDebInfo` is debatable, of course.

configure.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [[ $(uname -s) != Darwin ]]; then
+    cmake_options+=(
+        -DBUILD_SHARED_LIBS=OFF
+    )
+fi
```
masterleinad: What's that about?

configure.sh
```diff
@@ -0,0 +1,41 @@
...
+
+cmake -S "${source_dir}" -B "${build_dir}" "${cmake_options[@]}"
+make -C "${build_dir}" -j "$(computeProcsForMake)"
```
masterleinad:
```suggestion
cmake --build "${build_dir}" -j "$(computeProcsForMake)"
```
or just
```suggestion
cmake --build "${build_dir}" -j
```

configure.sh
```diff
@@ -0,0 +1,41 @@
...
+    else
+        nprocs=$(nproc)
+    fi
+
+    if (( nprocs > 3 )); then
+        echo $((nprocs - 2))
+    else
+        echo 1
+    fi
+}
```
masterleinad, line 20: Not sure if we need such an elaborate setup.

masterleinad (commented):
I really think users only need
```bash
source_dir=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
build_dir="${source_dir}/build"
cmake -S "${source_dir}" -B "${build_dir}"
cmake --build "${build_dir}" -j
```
but I might be biased.

configure.sh
g1257: What do you suggest to use?

configure.sh
g1257: -DCMAKE_COMPILE_WARNING_AS_ERROR should be OFF actually for users. I guess I'll just remove it because it must default to OFF, I imagine.

Release should be preferred for users.

configure.sh
g1257: Either gpt 5.6 or I hallucinated it. I'll remove.

configure.sh
masterleinad: Is it really a problem for this project to use the maximum hardware concurrency (which would imply using `-j` without arguments in the `CMake` build call)? On my M1, I'm seeing a peak memory usage of around 1.3 GB.

configure.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/bash
```
masterleinad: `/usr/bin/bash` doesn't exist for me. `/bin/bash` appears to be more standard.
```suggestion
#!/bin/bash
```

masterleinad (commented):
A `configure.sh` script normally doesn't invoke the "build" step. Would rather want to name it `build.sh`?

configure.sh
```diff
@@ -0,0 +1,32 @@
...
+source_dir=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+build_dir="${source_dir}/build"
+
+mkdir -p "${build_dir}"
+
+cmake_options=(
+    -DCMAKE_BUILD_TYPE=Release
+)
+
+cmake -S "${source_dir}" -B "${build_dir}" "${cmake_options[@]}"
```
masterleinad:
```suggestion
cmake -S "${source_dir}" -B "${build_dir}" -DCMAKE_BUILD_TYPE=Release
```

configure.sh
g1257: It is in virtual machines even if you don't oversubscribe. It freezes for a minute or so in some cases.

masterleinad (approved):

PR #264, opened 2026-08-14
Force KokkosKernels_ENABLE_TPL_BLAS/LAPACK

masterleinad:
As dicussed in Teams, it's easy for users to miss selecting a suitable TPL for BLAS and LAPACK in KokkosKernels. This pull request forces those two as TPLs when an external KokkosKernels installation couldn't be found. For an external KokkosKernels installation, the TPLs aren't yet exported at configuration but https://github.com/kokkos/kokkos-kernels/pull/3248 is supposed to fix that.

### DIFF
--- a/.github/workflows/ci-mac.yml
+++ b/.github/workflows/ci-mac.yml
@@ -29,9 +29,7 @@ jobs:
               -B ${{github.workspace}}/build \
               -GNinja \
               -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} \
-              -DCMAKE_COMPILE_WARNING_AS_ERROR=ON \
-              -DKokkosKernels_ENABLE_TPL_BLAS=ON \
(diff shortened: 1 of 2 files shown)


g1257 (approved):
I thought to set variables in the CMakeLists.txt was considered bad style; but I'm OK with this change.

masterleinad:
> I thought to set variables in the CMakeLists.txt was considered bad style; but I'm OK with this change.

The alternative would be to just error out but it's somewhat strange if `cmake ..` doesn't work at all.

g1257:
> > I thought to set variables in the CMakeLists.txt was considered bad style; but I'm OK with this change.
> 
> The alternative would be to just error out but it's somewhat strange if `cmake ..` doesn't work at all.

Yes, it'f fine. Let's not overthink it; it's a small change we can revert later if needed.

PR #271, opened 2026-08-19
Auto 4 sites and odd sites

g1257:
`@auto` not supported if sites < 5


dmrg/Engine/ParametersDmrgSolver.h
```diff
@@ -573,7 +575,8 @@ struct ParametersDmrgSolver {
 
 		std::cout << "FiniteLoopLengths=[";
 
-		if (lastSite == 1 && numberOfSites > 4) {
+		bool from_infinite = (numberOfSites < 5 && !isRestart);
```
PDoakORNL: I'd benefit from some documentation here I think, I'm not sure I can review this usefully as is.

Why do we know its "from infinite" if the number of sites is < 5 and isRestart is False?  I assume this is the @auto in the finite loop that is refferred to, but 72 looks like a test with four sites and @auto in the finite loop which contradicts the description of this PR so...

PDoakORNL (changes requested):
There needs to be some documentation or reference to some documentation, I don't know why this was wrong before and if it is actually correct now.

Perhaps your agent could generate some docs that you can check over for all this logic wrt to interaction between FiniteLoops and site count?

dmrg/Engine/ParametersDmrgSolver.h
g1257: I changed the logic there so small lattice don't need to be invoked. I've also documented the magic but only minimally.

dmrg/Engine/ParametersDmrgSolver.h
```diff
@@ -573,7 +575,9 @@ struct ParametersDmrgSolver {
 
 		std::cout << "FiniteLoopLengths=[";
 
-		if (lastSite == 1 && numberOfSites > 4) {
+		// For restarts and even (regular lattices), lastSite == 1 indicated
```
PDoakORNL, line 578: This is what I meant, enough local documentation like this if accurate will really help in the long term.

PDoakORNL (approved):
Not going to stand in the way of progress here.  Would still like to see us get more unit testing into dmrg.

PR #272, opened 2026-08-20
Fixes time evolution observations

g1257:
Removes exceptions from observe one-points and two-points, and instead returns 0 when data isn't available yet in the hd5 file.


PDoakORNL (commented):
So the obvious thing here is that there should be a unit test that protects this now correct behavior. The question is what work needs to be done to make that happen?  Is it just a question of Fake/Mock/Stubs needing to devised?

g1257:
Yes, something needs to be done as a unit test. I put it in draft for now.

PDoakORNL (approved):
LGTM =

PR #283, opened 2026-08-27
fixes `@auto` for all odd site lattices

g1257:
The case of lattice sites equals three was special; this PR leaves that case alone and fixes all other odd lattices.


PDoakORNL (approved):
Looks fine to me. What's the root cause of this and the previous fix?  Is this a indexing starts at 0 versus the methodology starts site counts at 1 or something more particular?

PR #284, opened 2026-08-28
GPUPlugin: Remove unused macros and functions

quantumsteve:
Remove some unset macros and unused functions in GPUPlugin. I also removed some unnecessary semicolons that were bugging me.


PDoakORNL (approved):
I see no problems here;

PR #289, opened 2026-09-01
Fix Kokkos dependency pairing in CMake

PDoakORNL:
## Summary

   Prevent CMake from combining a fetched Kokkos runtime with an installed
   KokkosKernels package built against a different Kokkos installation.  This happened to me 
   repeatedly with my own spack environments so we probably should cover this corner case,
   it seems like a user could hit it as well.

   ## Problem

   DMRG++ previously resolved Kokkos before KokkosKernels. If standalone Kokkos
   package discovery failed, CMake fetched Kokkos and created `Kokkos::kokkos`.
   A subsequent installed KokkosKernels package could then accept that existing
   target instead of loading the Kokkos installation against which it was built.

   This produced an executable containing both a fetched static Kokkos runtime and
   the shared Kokkos runtime required by KokkosKernels. The mixed Kokkos runtimes caused an 
   exit-time double-free during direct single-process execution. The failure was not reproduced under
   mpiexec -n 1, although the invalid runtime combination remained present.

   ## Changes

   - Probe for an installed KokkosKernels package before creating a standalone
     Kokkos target, allowing its package configuration to load its matching Kokkos.
   - If no installed pair is selected:
     - resolve Kokkos through the existing package/provider/FetchContent logic;
     - build KokkosKernels against that selected Kokkos target;
     - prevent an unrelated installed KokkosKernels package from being substituted.
   - Reject a pre-existing KokkosKernels target whose pairing cannot be verified.
   - Preserve the required BLAS and LAPACK options and installed-package checks.
   - Record Kokkos and KokkosKernels provenance in the CMake cache and configure
     output for diagnostics.
   - Preserve FetchContent providers, source overrides, disconnected operation,
     and fully fetched builds without hard-coded installation paths.

   ## Validation

   Validated the following configurations:

   - Installed KokkosKernels with its matching installed Kokkos.
   - Fully fetched Kokkos and KokkosKernels.
   - Installed Kokkos with source-built KokkosKernels.
   - Explicit KokkosKernels source override.

   For the installed pair, the Cincuenta link and runtime dependencies contain
   only the matching shared Kokkos/KokkosKernels installation; the fetched static
   Kokkos library is no longer present.

   For the fully fetched configuration, the executable uses only libraries from
   the corresponding FetchContent builds.

   The following checks also passed:

   - MPI-enabled Cincuenta build.
   - Direct `cincuenta -V`.
   - Direct `inputCincuenta0.ain` calculation.
   - `mpiexec -n 1` control run.
   - `ctest -R '^cincuenta0$'`.

This PR was prepared with the help of GPT-5.6-sol and terra.

### DIFF
--- a/PsimagLite/src/CMakeLists.txt
+++ b/PsimagLite/src/CMakeLists.txt
@@ -16,35 +16,43 @@ add_library(
   PsimagLite/Ainur/AinurSpirit.cpp
   PsimagLite/Ainur/AinurConvert.cpp)
 
-# Ensure Kokkos and KokkosKernels are available
+# Resolve Kokkos and KokkosKernels as a coherent pair. An installed
+# KokkosKernels package must be considered before a Kokkos target exists so its
+# package configuration can select the Kokkos installation it was built with.
 include(FetchContent)
-list(APPEND CMAKE_MESSAGE_INDENT "[Kokkos] ")
-FetchContent_Declare(
-  Kokkos
-  GIT_REPOSITORY https://github.com/kokkos/kokkos
-  GIT_TAG 5.2.0
-  SYSTEM FIND_PACKAGE_ARGS 5.1 NAMES Kokkos)
-FetchContent_MakeAvailable(Kokkos)
-list(POP_BACK CMAKE_MESSAGE_INDENT)
 
-list(APPEND CMAKE_MESSAGE_INDENT "[KokkosKernels] ")
-FetchContent_Declare(
-  KokkosKernels
-  GIT_REPOSITORY https://github.com/kokkos/kokkos-kernels
-  GIT_TAG 5.2.0
-  SYSTEM FIND_PACKAGE_ARGS 5.1 NAMES KokkosKernels)
-# Since DRMGPP requires BLAS and LAPACK, we should also make sure KokkosKernels uses these as TPLs
-set(KokkosKernels_ENABLE_TPL_BLAS
-    ON
-    CACHE BOOL "Whether to enable the BLAS TPL" FORCE)
-set(KokkosKernels_ENABLE_TPL_LAPACK
-    ON
-    CACHE BOOL "Whether to enable the LAPACK TPL" FORCE)
-FetchContent_MakeAvailable(KokkosKernels)
-if(NOT KokkosKernels_POPULATED)
+if(TARGET Kokkos::kokkoskernels)
+  message(
+    FATAL_ERROR
+      "A pre-existing Kokkos::kokkoskernels target cannot be verified as coherent with Kokkos::kokkos; let DMRG++ resolve the pair"
+  )
+endif()
+
+set(_dmrgpp_probe_kokkoskernels TRUE)
+if(FETCHCONTENT_TRY_FIND_PACKAGE_MODE STREQUAL "NEVER"
+   OR FETCHCONTENT_SOURCE_DIR_KOKKOS
+   OR FETCHCONTENT_SOURCE_DIR_KOKKOSKERNELS
+   OR TARGET Kokkos::kokkos)
+  set(_dmrgpp_probe_kokkoskernels FALSE)
+endif()
+
+if(_dmrgpp_probe_kokkoskernels)
+  list(APPEND CMAKE_MESSAGE_INDENT "[KokkosKernels] ")
+  find_package(KokkosKernels 5.1 QUIET CONFIG NAMES KokkosKernels)
+  list(POP_BACK CMAKE_MESSAGE_INDENT)
+endif()
+
+if(KokkosKernels_FOUND)
+  if(NOT TARGET Kokkos::kokkos OR NOT TARGET Kokkos::kokkoskernels)
+    message(FATAL_ERROR "The KokkosKernels package did not provide both Kokkos::kokkos and Kokkos::kokkoskernels")
+  endif()
+  set(_dmrgpp_kokkos_provenance "package via KokkosKernels")
+  set(_dmrgpp_kokkoskernels_provenance "package")
+
+  # Remove values forced for a source build by an earlier configure. Newer
+  # installed packages export these as normal variables that can be checked.
   unset(KokkosKernels_ENABLE_TPL_BLAS CACHE)
   unset(KokkosKernels_ENABLE_TPL_LAPACK CACHE)
-  # Past version 5.2, KokkosKernels exports its TPLs as non-CACHE variables
   if(KokkosKernels_VERSION VERSION_GREATER_EQUAL 5.2.99)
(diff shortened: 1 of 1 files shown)


PsimagLite/src/CMakeLists.txt
```diff
@@ -16,35 +16,43 @@ add_library(
...
-    CACHE BOOL "Whether to enable the LAPACK TPL" FORCE)
-FetchContent_MakeAvailable(KokkosKernels)
-if(NOT KokkosKernels_POPULATED)
+
+if(TARGET Kokkos::kokkoskernels)
+  message(
+    FATAL_ERROR
+      "A pre-existing Kokkos::kokkoskernels target cannot be verified as coherent with Kokkos::kokkos; let DMRG++ resolve the pair"
+  )
+endif()
```
masterleinad, line 29: What is the user supposed to do in that case? Can't we just trust that a previously found `Kokkos::kokkoskernels` installation is fine? What's the risk there? If `Kokkos::kokkoskernels` is a target then `Kokkos::kokkos` should also exist as a target.

PsimagLite/src/CMakeLists.txt
```diff
@@ -16,35 +16,43 @@ add_library(
...
+  if(NOT TARGET Kokkos::kokkos OR NOT TARGET Kokkos::kokkoskernels)
+    message(FATAL_ERROR "The KokkosKernels package did not provide both Kokkos::kokkos and Kokkos::kokkoskernels")
+  endif()
+  set(_dmrgpp_kokkos_provenance "package via KokkosKernels")
+  set(_dmrgpp_kokkoskernels_provenance "package")
+
+  # Remove values forced for a source build by an earlier configure. Newer
+  # installed packages export these as normal variables that can be checked.
   unset(KokkosKernels_ENABLE_TPL_BLAS CACHE)
   unset(KokkosKernels_ENABLE_TPL_LAPACK CACHE)
```
masterleinad, line 55: We unset these variables before we set them below? Is that what we intent to do on the first configuration run?

masterleinad (commented):
Would it be sufficient to just not link against `Kokkos::kokkos`? That should at least make the surface for the issue smaller.

PsimagLite/src/CMakeLists.txt
PDoakORNL: These lines predate this PR. They remove cache options used to configure a source-built KokkosKernels before inspecting an installed package’s exported feature variables. This prevents stale forced ON values from being mistaken for properties of the installed library, particularly when reconfiguring an existing build directory from fetched to installed dependencies.

They are less central now that source options are set only in the fallback branch, but retaining them safely handles stale CMake caches.

PsimagLite/src/CMakeLists.txt
PDoakORNL: Well I think the installation usually is fine but if your Kokkos::kokkoskernels is coming from a spack view with default.link = roots and you search for Kokkos first then Kokkos::kokkos is not visible when the cmake earilier looks for it and it gets fetched. This isn't the default link method but in this environment I need it because I have other packages with conflicting tool chains.  If we don't find Kokkos on its own then even in this case this doesn't happen.

We can use the following policy and lose most of the complication of the initial version of this PR.

1. Prefer installed KokkosKernels and let its package configuration select matching Kokkos.
2. Otherwise fetch both Kokkos and KokkosKernels.
3. Link PsimagLite only to Kokkos::kokkoskernels.
4. Reject pre-existing Kokkos targets whose pairing cannot be verified.

This should work in all cases.

masterleinad: That sounds reasonable

PsimagLite/src/CMakeLists.txt
```diff
@@ -16,44 +16,32 @@ add_library(
...
+if(TARGET Kokkos::kokkos OR TARGET Kokkos::kokkoskernels)
+  message(FATAL_ERROR "Kokkos targets must not exist before DMRG++ resolves Kokkos and KokkosKernels as a pair")
+endif()
+
+set(_dmrgpp_kokkoskernels_package_found FALSE)
+if(NOT FETCHCONTENT_TRY_FIND_PACKAGE_MODE STREQUAL "NEVER"
+   AND NOT FETCHCONTENT_SOURCE_DIR_KOKKOS
+   AND NOT FETCHCONTENT_SOURCE_DIR_KOKKOSKERNELS)
+  list(APPEND CMAKE_MESSAGE_INDENT "[KokkosKernels] ")
+  find_package(KokkosKernels 5.1 QUIET CONFIG NAMES KokkosKernels)
```
masterleinad, line 33:
```suggestion
  find_package(KokkosKernels 5.2 QUIET CONFIG NAMES KokkosKernels)
```
while we are at it

PsimagLite/src/CMakeLists.txt
masterleinad: Fair enough.

PsimagLite/src/CMakeLists.txt
```diff
@@ -62,10 +50,56 @@ if(NOT KokkosKernels_POPULATED)
       message(FATAL_ERROR "KokkosKernels must be configured with LAPACK support!")
     endif()
   endif()
+else()
+  # Do not substitute either source dependency with an independently installed
+  # package; the two source builds must use the same Kokkos target.
+  block()
```
masterleinad, line 56: Csn you clarify whey we need/want `block` here?

PDoakORNL: that makes all the variable definitions until endblock() scoped in that range.  The purpose being that (FETCHCONTENT_TRY_FIND_PACKAGE_MODE NEVER) is not something we want leaking out globally we just want that option here to force a fetch of the packages together and not let cmake try and find them.  it had it's chance to find KokkosKernels and didn't.

PDoakORNL (commented):
bumped version, explained what block() is doing.

PsimagLite/src/CMakeLists.txt
masterleinad: My concern is that `block` prevents passing extra `CMake` configuration option to `KokkosKernels` (like `KokkosKernel_ENABLE_TPL_ROCBLAS=ON`) but I haven't investigated if that's the case or not.

PDoakORNL: My understanding is it creates a local scope that can shadow but is not clean.  The other options should come through.

masterleinad (approved):

PsimagLite/src/CMakeLists.txt
PDoakORNL: Verified cmake options for KokkosKernel and Kokkos make it into the build even when the path is through the block.

PR #290, opened 2026-09-02
Liouvillian with "Squaring" Method

g1257:
Introduces SolverOptions "LdaggerL", which causes  L|x> = 0 to be solved for |x> where L is the vectorized Liouvillian by keeping L and L* in CRS in memory, and doing for every Lanczos vector y, L*(Ly), so that the construction is Hermitian. This is better than the dense inversion we were using but it's limited by the amount of RAM used for storing L and L^* in CRS.

_TODO_:  Create a way to test labels other than lowest energy (**), like <gs|sz_p|gs> and <gs|sz|gs>. Then test these two labels for input8902 and input8903.
I suggest this _TODO_ item be done in a separate PR.

(*) A following PR will add support for  "LdaggerL" when MatrixVectorKron, whereas   "LdaggerL" when MatrixVectorOnTheFly will remain forbidden, because it's not worth the effort.

(**) A following (recommended) or preceeding PR will implement this feature.


dmrg/Engine/tests/test_LdaggerL.cpp
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	~TemporaryDirectory() { std::filesystem::remove_all(path_); }
+
+	const std::filesystem::path& path() const { return path_; }
+
+private:
+
+	std::filesystem::path path_;
+};
```
masterleinad, line 40: Do you actually need a somewhat random directory name? Can't you just use a subdirectory in the current build directory?

dmrg/Engine/tests/test_LdaggerL.cpp
```diff
@@ -0,0 +1,179 @@
...
+	const std::string           command = "cd " + shellQuote(workDir.string()) + " && "
+	    + shellQuote(DMRG_EXECUTABLE_PATH) + " -f " + shellQuote(input.string()) + " > "
+	    + shellQuote(output.string()) + " 2>&1";
+	const int status = std::system(command.c_str());
+
+	std::ifstream      stream(output);
+	std::ostringstream buffer;
+	buffer << stream.rdbuf();
+	return { status, buffer.str() };
+}
```
masterleinad, line 67: I'm not sure if it's best for this test to be purely in C++. It might be easier to just execute `dmrg` directly and intrpspect the output with other bash tools like we do for other tests.

masterleinad (commented):
What's the motivation for executing dmrg via a system call from a different C++ executable instead of running it via CTest directly and introspecting the output with command-line tools?

dmrg/Engine/MatrixVectorStored.h
```diff
@@ -114,7 +114,13 @@ class MatrixVectorStored : public MatrixVectorBase<ComplexOrRealType_> {
 		const bool         debugMatrix = options.isSet("debugmatrix");
 
 		matrixStored_.clear();
+
 		hc.fullHamiltonian(matrixStored_, aux, model.isHermitian());
+		const bool ldaggerL = model.params().options.isSet("LdaggerL");
+		if (ldaggerL) {
+			transposeConjugate(transpose_, matrixStored_);
+		}
+
```
PDoakORNL, line 125: A unit test of the constructor would check to see that the invariant that the I guess matrix in storage would be transposed here.  That would be your regression test of MatrixVectorStored behavior when LdaggerL isSet.  It could also check that when LdaggerL is not set this doesn't happen.

dmrg/Engine/MatrixVectorStored.h
```diff
@@ -114,7 +114,13 @@ class MatrixVectorStored : public MatrixVectorBase<ComplexOrRealType_> {
 		const bool         debugMatrix = options.isSet("debugmatrix");
 
 		matrixStored_.clear();
+
 		hc.fullHamiltonian(matrixStored_, aux, model.isHermitian());
+		const bool ldaggerL = model.params().options.isSet("LdaggerL");
```
PDoakORNL, line 121: Since this boolean changes the behavior of the object and is immutable after construction this should really be a data member.  You recalculate it later which is a bad smell.
Since model exists at construction this member variable can go into the initializer list as well.
```
MatrixVectorStored(const ModelType&                     model,
	                   const HamiltonianConnectionType&     hc,
	                   const typename ModelHelperType::Aux& aux)
	    : model_(model),
          isLDaggerL_(model.params().options.isSet("LdaggerL"),
	    , progress_("MatrixVectorStored")
	{

File: dmrg/Engine/MatrixVectorStored.h [hunk 5]
```diff
@@ -137,6 +143,12 @@ class MatrixVectorStored : public MatrixVectorBase<ComplexOrRealType_> {
 	void matrixVectorProduct(SomeVectorType& x, SomeVectorType const& y) const
 	{
 		matrixStored_.matrixVectorProduct(x, y);
+		const bool ldaggerL = model_.params().options.isSet("LdaggerL");
```
PDoakORNL, line 143: Don't recalculate this construction time invariant

dmrg/Engine/tests/test_LdaggerL.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+RunResult runDmrg(const std::filesystem::path& input, const std::filesystem::path& workDir)
+{
+	const std::filesystem::path output  = workDir / "process-output.txt";
+	const std::string           command = "cd " + shellQuote(workDir.string()) + " && "
+	    + shellQuote(DMRG_EXECUTABLE_PATH) + " -f " + shellQuote(input.string()) + " > "
+	    + shellQuote(output.string()) + " 2>&1";
+	const int status = std::system(command.c_str());
+
```
PDoakORNL, line 62: The "right" thing to do here wrt unit testing is just construct the DmrgRunner which is what your testing.  We don't need to test main or write all this shell escaping, then you won't have to write a file.  But see following comments I don't think this requires changes to DmrgRunner

dmrg/Engine/DmrgRunner.cpp
```diff
@@ -129,6 +129,8 @@ void DmrgRunner<RealType>::doOneRun3(const OptionsForIntrospect& op_options) con
 	//! Setup the Model
 	ModelSelector<ModelBaseType> modelSelector(dmrg_solver_params_->model);
 	ModelBaseType&               model = modelSelector(*dmrg_solver_params_, *io_, geometry);
+	if (dmrg_solver_params_->options.isSet("LdaggerL") && model.isHermitian())
+		throw PsimagLite::RuntimeError("LdaggerL cannot be used with a Hermitian model\n");
```
PDoakORNL, line 126: seems like a low level input error to be trapped in the runner seems like the model factory should do this. Then you don't have to unit test this high level class.

PDoakORNL (changes requested):
I don't know if Hermes/GPT helped you here but this sort of putting the logic into the wrong object or layer is a frequent failure path for gpt-sol its annoying.  The shell call into catch macros also smells like reward hacking for "write me Catch2 unit tests of this"

g1257 (commented):
I think I've addressed all your points. Please see what you think now. Thanks.

dmrg/Engine/MatrixVectorBase.h
```diff
@@ -86,11 +86,12 @@ DISCLOSED WOULD NOT INFRINGE PRIVATELY OWNED RIGHTS.
 #include <vector>
 
 namespace Dmrg {
-template <typename ComplexOrRealType_> class MatrixVectorBase {
+template <typename ComplexOrRealType_, typename TypesType_ = MatrixVectorTypes<ComplexOrRealType_>>
```
masterleinad: Where are we using it with a different template argument than the default?

dmrg/CMakeLists.txt
```diff
@@ -310,7 +310,61 @@ add_test(NAME input6020 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input6020.ain) # ene
...
-add_test(NAME input8901 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input8901.ain "<gs|sz|gs>") # none
+add_test(NAME input8901 COMMAND $<TARGET_FILE:dmrg> -f ${PATH_TO_INPUTS}/input8901.ain "<gs|sz|gs>") # none
+set_tests_properties(input8901 PROPERTIES WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/input8901)
+
+set(_ldaggerl_input_text)
+file(READ ${PATH_TO_INPUTS}/input8902.ain _ldaggerl_input_text)
+string(FIND "${_ldaggerl_input_text}" "matrixvectorstored," _ldaggerl_stored_position)
+if(_ldaggerl_stored_position EQUAL -1)
+  message(FATAL_ERROR "input8902.ain lacks matrixvectorstored")
+endif()
```
masterleinad: The input file is static so it feels a little weird to test it here apart from documenting a requirement.

dmrg/CMakeLists.txt
```diff
@@ -310,7 +310,61 @@ add_test(NAME input6020 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input6020.ain) # ene
...
+
+string(REPLACE "matrixvectorstored," "MatrixVectorOnTheFly," _ldaggerl_onthefly_input "${_ldaggerl_input_text}")
+set(_ldaggerl_onthefly_input_file ${CMAKE_CURRENT_BINARY_DIR}/input8902_ldaggerl_onthefly.ain)
+file(WRITE ${_ldaggerl_onthefly_input_file} "${_ldaggerl_onthefly_input}")
+
+set(_isolated_dmrg_test_names input8901 input8902 input8903 ldaggerl_hermitian_rejected ldaggerl_kron_rejected
+                              ldaggerl_onthefly_rejected)
+foreach(test_name IN LISTS _isolated_dmrg_test_names)
+  file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${test_name})
+endforeach()
```
masterleinad: If you add the tests first, you can do
```suggestion
foreach(test_name IN LISTS _isolated_dmrg_test_names)
  file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${test_name})
  set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${test_name})
endforeach()
```
instead of repeating yourself for every test.

dmrg/CMakeLists.txt
g1257: Removed.

dmrg/CMakeLists.txt
```diff
@@ -310,7 +310,52 @@ add_test(NAME input6020 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input6020.ain) # ene
 add_lowest_eigenvalue_check_test(output6020 input6020 -2.59474 runForinput6020.cout)
 add_test(NAME input6700 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input6700.ain) # energy -4.17889 we don't check this
 add_test(NAME input8900 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input8900.ain "<gs|sz_p|gs>") # none
-add_test(NAME input8901 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input8901.ain "<gs|sz|gs>") # none
+add_test(NAME input8901 COMMAND $<TARGET_FILE:dmrg> -f ${PATH_TO_INPUTS}/input8901.ain "<gs|sz|gs>") # none
+set_tests_properties(input8901 PROPERTIES WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/input8901)
+
+set(_ldaggerl_input_text)
+file(READ ${PATH_TO_INPUTS}/input8902.ain _ldaggerl_input_text)
+string(FIND "${_ldaggerl_input_text}" "matrixvectorstored," _ldaggerl_stored_position)
```
masterleinad: This is unused now.
```suggestion
```

PDoakORNL (approved):
Ok this looks good to me now.

PR #291, opened 2026-09-02
MatrixVector PR Series Part I: Simplify template

g1257:
This is part of a PR series to make MatrixVector hierarchy go from templated to virtual inheritance. It'll lower compilation times but also it'll make it easier to fit planned features.

Part I: Changes the ModelType template of ``MatrixVectorBase<ModelType>`` into ``MatrixVectorBase<ComplexOrRealType>`` (and for all derived classes).

Note that the intention is for ComplexOrRealType to take only double or std::complex<double>, but that's not enforced with a static_assert or anything. In any case, executables like dmrg and observe and cincuenta do instantiate both types and the exec. path is chosen at runtime from the input.

### DIFF
--- a/dmrg/Engine/DmrgRunner.cpp
+++ b/dmrg/Engine/DmrgRunner.cpp
@@ -94,7 +94,7 @@ void DmrgRunner<RealType>::doOneRun2(const OptionsForIntrospect& op_options) con
 	                                         InputNgType::Readable,
 	                                         SuperGeometryType>;
 	using MatrixVectorModelType  = typename MatrixVectorTypes<ComplexOrRealType>::ModelType;
-	static_assert(std::is_same<ModelBaseType, MatrixVectorModelType>::value,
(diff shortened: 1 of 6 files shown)


dmrg/Engine/DmrgRunner.cpp
```diff
@@ -91,14 +93,17 @@ void DmrgRunner<RealType>::doOneRun2(const OptionsForIntrospect& op_options) con
 	                                         ParametersDmrgSolverType,
 	                                         InputNgType::Readable,
 	                                         SuperGeometryType>;
+	using MatrixVectorModelType  = typename MatrixVectorTypes<ComplexOrRealType>::ModelType;
+	static_assert(std::is_same<ModelBaseType, MatrixVectorModelType>::value,
```
quantumsteve:
```suggestion
	static_assert(std::is_same_v<ModelBaseType, MatrixVectorModelType>,
```

quantumsteve (approved):
Just one minor style recommendation.

PR #292, opened 2026-09-04
MatrixVector PR Series Part II: select implementation virtually

g1257:
Use a scalar-templated owning facade to select Stored, OnTheFly, or Kron at runtime while keeping sector-local lifetimes intact. Define the virtual base contract and collapse DmrgRunner dispatch to one solver graph per scalar type.

I'm planning a part III where I'll delete some unused files that refer to MatrixVector hierarchy but were obsoleted even before this PR.


masterleinad (commented):
Have you measured any improvements with this pull request?

masterleinad (approved):

PR #298, opened 2026-09-08
Check Observables ("labels") insitu

g1257:
DMRG++ can be run as ./dmrg -f input.ain "insitu_observation_here"; This MR enables regression tests for the insitu observation written to the .cout. This MR uses TextAndNumbersCheck.cpp instead of the check_value used for lowest energy.


PsimagLite/examples/checkTextAndNumbers.cpp
```diff
@@ -41,25 +64,26 @@ int main(int argc, char* argv[])
...
+		const PsimagLite::TextAndNumbersChecker checker(parseTolerance(argv[3]),
+		                                                ignoredLinePrefix);
+
+		if (option.mode == ComparisonMode::Observable)
+			checker.runObservables(argv[1], argv[2], option.value);
+		else
+			checker.run(argv[1], argv[2]);
 	} catch (const std::exception& error) {
 		std::cerr << error.what() << '\n';
 		return EXIT_FAILURE;
```
PDoakORNL, line 87: This program is not trivial and this directory should be named tests, but I think the answer is to convert all the test programs in examples/ to Catch2 programs at which point this probably isn't needed.

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
```diff
@@ -206,22 +276,181 @@ class TextAndNumbersChecker {
...
+
+	//---------------------------------------------------------------------------//
+	/*!
+	 * \brief Recognize strict decimal or exponent real-number syntax
+	 */
+	static bool isDoubleSyntax(const std::string& token)
+	{
+		if (token.empty())
+			return false;
+
```
PDoakORNL, line 396: This seems like so much code caused by the need to output to string to test most PsimagLite objects.  I don't think coverage of the operator<< is as crucial as native numerical testing.  This may be some of the most extensive testing of iostreams I've ever seen.  The only thing I see justifying this long term is if we need to interoperate with consumers that can not take binary hdf5 input for numerics and I kind of this we shouldn't be doing that.

For users who like text files this is super strong support but...

PsimagLite/src/tests/testCheckTextAndNumbers.cpp
```diff
@@ -382,4 +410,330 @@ TEST_CASE("TextAndNumbersChecker compares real numbers with absolute tolerance",
 	}
 }
 
+TEST_CASE("TextAndNumbersChecker compares complex numbers by magnitude", "[TextAndNumbersChecker]")
+{
```
PDoakORNL, line 414: This is a pretty sweet test program though, the input seems like it could be hiding corner cases especially wrt to mixed precision etc.  But not in this PR

PsimagLite/examples/CMakeLists.txt
```diff
@@ -29,6 +29,17 @@ target_link_libraries(checkTextAndNumbers psimaglite)
...
+set(check_text_observable_actual ${CMAKE_CURRENT_BINARY_DIR}/checkTextAndNumbers_observable_actual.txt)
+file(WRITE ${check_text_observable_reference} "5 (9,9) 0.3 <gs|sz|P0> (8,8)\n5 (-0.18,0.05) 0.3 <gs|sz|P0> (0,0)\n")
+file(
+  WRITE ${check_text_observable_actual}
+  "PsiApp: CmdLine: dmrg -f input21.ain <gs|sz|P0>\n5 (-7,-7) 0.300000 <gs|sz|P0> (-6,-6)\n5 (-0.18,0.05) 0.300000 <gs|sz|P0> (0,0)\n"
+)
+add_test(NAME checkTextAndNumbers_observables
+         COMMAND $<TARGET_FILE:checkTextAndNumbers> ${check_text_observable_reference} ${check_text_observable_actual}
+                 1e-6 "--insitu-label=<gs|sz|P0>")
+
```
PDoakORNL, line 42: But don't we really want to check this against a byte perfect binary reference?

PDoakORNL (approved):

PsimagLite/examples/checkTextAndNumbers.cpp
g1257: TextAndNumbersChecker.cpp has both an example (this one) and massive Catch2 unit tests.

PsimagLite/examples/CMakeLists.txt
g1257: "Iin-situ" measurements aren't written to the hdf5, only to the .cout. They should definitely go to the hdf5; https://github.com/dmrgpp-project/dmrgpp/issues/302

PsimagLite/src/PsimagLite/TextAndNumbersChecker.hpp
g1257: gpt 5.6 sol went overboard here without even prompting as we discussed.

PR #303, opened 2026-09-10
Track physical time for P vectors: Part I

g1257:
Propagate per-vector time through time evolution, temporary vectors, and compatible sums. Add focused Catch2 coverage for default, assigned, and summed times.


dmrg/Engine/Pvector.h
```diff
@@ -40,6 +41,8 @@ template <typename ComplexOrRealType> class Pvector {
 		const SizeType n = vStr_.size();
 		if (n == 0 || vStr_[n - 1] != "DONE" || other.vStr_.size() == 0)
 			err("Pvector::sum\n");
+		if (time_ != other.time_)
```
PDoakORNL, line 44: Floating point times that are equivalent may not be exact here, can we count on that?

dmrg/Engine/Pvectors.h
```diff
@@ -149,6 +157,9 @@ template <typename TargetingBaseType> class Pvectors {
 	                 PsimagLite::String       p0PlusP1)
 	{
 		assert(ind0 < ind1);
+		if (pVectors_[ind0]->time() != pVectors_[ind1]->time())
```
PDoakORNL, line 160: Same floating point time comparison

dmrg/Engine/NonLocalForTargetingExpression.h
```diff
@@ -127,6 +127,15 @@ template <typename TargetingBaseType> class NonLocalForTargetingExpression {
 		    block1,
 		    isLastCall);
 
+		const VectorSizeType& indices = oneTimeEvolution->indices();
+		assert(indices.size() == timeParams.timeSteps);
+		const RealType baseTime = oneTimeEvolution->time();
+		for (SizeType i = 0; i < indices.size(); ++i) {
+			// Keep this offset synchronized with Pvectors::initTimeVectors().
+			const RealType offset = i * timeParams.tau / (timeParams.timeSteps - 1);
```
PDoakORNL, line 135: what if timeParams.timeSteps=1 is that protected?

dmrg/Engine/Pvectors.h
```diff
@@ -86,12 +86,20 @@ template <typename TargetingBaseType> class Pvectors {
...
+		pVectors_[ind]->setTime(time);
+	}
+
 	template <typename SomeLambdaType>
-	void createNew(const VectorWithOffsetType& src, SomeLambdaType& lambda)
+	void createNew(const VectorWithOffsetType& src, RealType time, SomeLambdaType& lambda)
 	{
 		const SizeType           ind   = aoeNonConst().createPvector(src);
 		const PsimagLite::String ename = lambda(ind);
 		PvectorType*             pnew  = new PvectorType(ename);
```
PDoakORNL, line 101: this pattern looks like a missing constructor, but as I read it setTime isn't just used to finish a valid construction.

PDoakORNL (changes requested):
I've got to pack it in for the day but some intial questions which may be addressed in the following PR's, and this is with some help from my robot so I may have not read carefully enough:

What happens to pvectors on restart they don't seem to serialize their time just their vector?  Seems like the pvectors own time_ doesn't get serialized with it.  consequently it can't be read in if that is something that is desirable.

I think if we assign one pvector to another its time_ doesn't come with it should their be something like 
```
pvectors_.setTime(i, pvectors_(x).time());
```
after
```
v0 = this->tv(x);
v0 *= tmp.term(0).ket().factor();
```
around TargetingExpression.h:327

PDoakORNL:
It does seem like your stack PRs are aimed at the unanchored comments so I think what there is to address is just the in source feedback.

dmrg/Engine/NonLocalForTargetingExpression.h
g1257: My suggestion is as follows.

(1) For this PR: Assert in this location and a throw at the time of reading from input file.

(2) New PR where timeSteps > 1 is properly enforced throughout.

Please let me know what you think.

dmrg/Engine/Pvector.h
g1257: It works as the only time arithmetic  seems to be adding tau many times.

What do you suggest? Using an epsilon?

PDoakORNL: In general a floating point comparison should have some epsilon how big it should be really depends on how different a time could be. In practice I have of course written such things too and gotten questioned in review.

I thought about this on my drive and If its only addition of a constant tau and never any other operation on a time this should be fine, but a time carried between restarts that is not getting passed by simple bit/bytewise  read and write would be brittle. There could also be some portability issues for times coming from a previous run even then depending on how different the build or architecture was that produced the data. Its been awhile since i looked at comparisons and the floating point standards they give some guarantees but they aren't very strong. 

So we can keep this but it introduces a very strong constraint about times wrt pvectors which I'd document at the class level as such, maybe at the declaration of `time_`

dmrg/Engine/NonLocalForTargetingExpression.h
PDoakORNL: That sounds good enough.

dmrg/Engine/Pvector.h
masterleinad: I highly recommend using a tolerance here. AFAICT, there is no guarantee that `tau` is constant.

masterleinad: We might even consider testing with `-Wfloat-equal`.

dmrg/Engine/Pvectors.h
g1257: I don't understand.

(1) Pvector::setTime() is a trivial setter

(2) Pvectors::createNew  can become a void Pvectors::push(const Pvector& pvector) perhaps but not a ctor for Pvectors (maybe for Pvector)

Anyway, let me know if you prefer changes here.

PDoakORNL: To Clarify I'm talking about the Pvector constructor.  There is nothing wrong with the setter I prefer it to direct access to the variable. I think createNew is fine unless you want to express this has push semantics.
I like the principle that an object should be valid when the constructor completes. Constructors establish the arguments required create a valid object instance as well as any details about the ordering of operations etc in the object itself instead of at the site of the objects creation. This enhances encapsulation, make its easier to reason locally about the object, make it easier to write tests, etc. 

So here I see one way a pvector can be made a valid object is construction with a name + setTime + setAsDone. This seems like it belongs in Pvector and not Pvectors.

PDoakORNL (approved):
I don't want to bike shed this to death the issues that relate to progressing the feature are addressed.

PR #307, opened 2026-09-15
Psimaglite: replace sparse matrix multiply

g1257:
Replace the provenance-blocked routine with an independently authored row-local sparse accumulator. Validate dimensions, preserve output aliasing, sort result columns, and discard exact cancellation zeros. Add focused real, complex, randomized, aliasing, and edge-case tests.


PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
-					temp[ptr[jbk]]  = tmp;
-					index[ptr[jbk]] = jbk;
-					itemp++;
-				} else {
-					temp[ptr[jbk]] += tmp;
-				}
+	if (A.cols() != B.rows())
+		throw RuntimeError("CrsMatrix::multiply: incompatible matrix dimensions\n");
+
+	CrsMatrix<S> result;
```
masterleinad: Why are you doing an extra allocation instead of working on `C` like before?

masterleinad (approved):
It might be interesting to profile this implementation to see if `unordered_map` is a good choice here and if sorting is beneficial. So far I haven't seen this function pop-up in any profilles, though.

PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
-				tmp = A.getValue(j) * B.getValue(k);
-				if (ptr[jbk] < 0) {
-					ptr[jbk]        = itemp;
-					temp[ptr[jbk]]  = tmp;
-					index[ptr[jbk]] = jbk;
-					itemp++;
-				} else {
-					temp[ptr[jbk]] += tmp;
-				}
+	if (A.cols() != B.rows())
```
PDoakORNL, line 986: This isn't free although it might be nearly so if nothing else is putting pressure on branch prediction, do we want to do this new check in release or would this only happen due to what I'd call a development time defect and so a debug only assert would be better?

PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
+		throw RuntimeError("CrsMatrix::multiply: incompatible matrix dimensions\n");
+
+	CrsMatrix<S> result;
+	result.resize(A.rows(), B.cols());
```
PDoakORNL, line 990: What C's dimensions are when it comes in could have a big impact here. If C's resize didn't actually induce reallocation here then result could cause cause a performance delta. If on the other hand C usually did have to reallocate than think swap at the end will make this all practically the same cost.

PDoakORNL (approved):
I think we should watch out for a performance regression here and I'd personally not keep the new idiot check for dimensions in release. Although reflexively not bounds checking is maybe a C/C++ instinct that isn't always great.

PsimagLite/src/PsimagLite/CrsMatrix.h
masterleinad: I wouldn't be concerned about this one check. Everything else we do here is much more expensive.

PsimagLite/src/PsimagLite/CrsMatrix.h
masterleinad: `resize` effectively only does
```
rowptr_.resize(nrow + 1);
```
so is fairly cheap.

PsimagLite/src/PsimagLite/CrsMatrix.h
g1257: We could do an assert. Just to be clear this check isn't in a loop. I don't think this function is ever inlined, so I can't believe there's branch prediction at play here.

PsimagLite/src/PsimagLite/CrsMatrix.h
g1257: Doing a quick inspection I see that C comes with its component vectors (rowptr, colind, values) all empty, I mean, size 0.

Note also that CrsMatrix::resize() with two arguments doesn't do resize semantics, it clears all first. That's bad, but to be properly inconsistent, the CrsMatrix::resize() with 3 args. does have resize semantics.

g1257:
> It might be interesting to profile this implementation to see if `unordered_map` is a good choice here and if sorting is beneficial. So far I haven't seen this function pop-up in any profilles, though.

Agreed. FWIW, a quick search of "git grep multiply" yields call sites (1) under directory models, and (2) for observe. For models the sizes are 2, 4, 16, and 64 at most. The observe calls are in the thousands or more. So this change may or may not have introduced a performance regression in observe, but not in dmrg.

C.resize(rows, cols) doesn't resize the values and cols because it doesn't have the number of non-zeroes passed in. I suspect C.pushValue and C.pushColumn are expensive as there's no reserve beforehand. This pattern appears in both the previous implementation and this one.

masterleinad (approved):

PR #308, opened 2026-09-15
dmrg: replace permutation parity implementation

g1257:
This PR replaces dmrg's permutation parity implementation with an independently authored inversion-parity helper. Add deterministic coverage for identities, cycles, reversals, large permutations, random transpositions, and inverse permutations.


dmrg/Engine/PermutationParity.h
```diff
@@ -0,0 +1,22 @@
+#ifndef DMRG_PERMUTATION_PARITY_H
+#define DMRG_PERMUTATION_PARITY_H
+
+#include <PsimagLite/Vector.h>
+#include <vector>
+
+namespace Dmrg {
+
+inline int parityOfPermutation(const std::vector<SizeType>& permutation) noexcept
```
PDoakORNL, line 9: Without a micro benchmark its hard to say if this is any more or less efficient than the exiting implementation, to me it doesn't look grossly different in terms of loop struture.  I certainly can't head eval all the bitshift stuff in the original so this has the huge advantage of being human readable.

PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
-				tmp = A.getValue(j) * B.getValue(k);
-				if (ptr[jbk] < 0) {
-					ptr[jbk]        = itemp;
-					temp[ptr[jbk]]  = tmp;
-					index[ptr[jbk]] = jbk;
-					itemp++;
-				} else {
-					temp[ptr[jbk]] += tmp;
-				}
+	if (A.cols() != B.rows())
```
PDoakORNL: This isn't free it might be nearly so if nothing else is putting pressure on branch prediction, do we want to do this new check in release or would this only happen due to what I'd call a development time defect and so a debug only assert would be better?

PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
+		throw RuntimeError("CrsMatrix::multiply: incompatible matrix dimensions\n");
+
+	CrsMatrix<S> result;
```
PDoakORNL: So this is new, we used to just use C inplace.

PsimagLite/src/PsimagLite/CrsMatrix.h
```diff
@@ -978,52 +980,53 @@ void printFullMatrix(const CrsMatrix<T>& s,
...
+	result.resize(A.rows(), B.cols());
```
PDoakORNL: What C is when it comes in could have a big impact here.  Assuming the optimizer doesn't do anything cool, If C's resize didn't actually induce reallocation here then result could cause cause a performance delta.  If on the other hand C usually did have to reallocate than think swap at the end will make this all practically the same cost.

PDoakORNL (approved):
I feel like broadly it would be good if we had a way to plot the execution time of the application tests that aren't trivial and just catch any gross regression we might cause.  At the moment I guess @masterleinad  or @quantumsteve will probably notice, or are we so focused on the hot GPU region there that CPU side changes could slip by?

PR #310, opened 2026-09-17
Track physical time for P vectors: Part IV

g1257:
Temporary vectors need time tracking as well and there was a wrong fall through after vector computed or "DONE".


thomas-jinu (commented):
Hi Gonzalo; 

It worked for "input....apply" but not for "input....evolve" 

It reintroduced the error of not finding the Hamiltonian value on table for "input....evolve" ; meaning the times associated with Pvectors are not tracked in the restart from the apply stage.

PR #312, opened 2026-09-18
V7.01 release

g1257:
When this PR gets merged DMRG++ will be tagged as v7.01 as a new release.


dmrg/CMakeLists.txt
```diff
@@ -226,6 +231,9 @@ add_test(NAME input73 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input73.ain) # energy
 add_lowest_eigenvalue_check_test(output73 input73 -4.47214 runForinput73.cout)
 add_test(NAME input74 COMMAND ./dmrg -f ${PATH_TO_INPUTS}/input74.ain) # energy -0.707712
 add_lowest_eigenvalue_check_test(output74 input74 -0.707712 runForinput74.cout)
+add_test(NAME input74_dmrg_version COMMAND ${H5DUMP_EXECUTABLE} -r -d /Def/ApplicationInfo/DmrgVersion
+                                           ${CMAKE_CURRENT_BINARY_DIR}/data74.hd5)
+set_tests_properties(input74_dmrg_version PROPERTIES DEPENDS input74 PASS_REGULAR_EXPRESSION "\"v7.01\"")
```
PDoakORNL: I'd like all the version info to go back to a single source of truth, which we should consider to be?  I'd like the last version tag in history of the current branch.  But this doesn't work if the code is distributed as a tarball. So..

PDoakORNL (commented):
From how this looks we need to update the version output from cincuenta as well. or basically all the executables should at least be able to report the version number and probably put the version number in there output if they don't actually flow through this dmrg path.

dmrg/CMakeLists.txt
g1257: Version.h contains "7.1" in a centralized way.
Note that
```
cd build
cincuenta/src/cincuenta -V
```
yields v7.1 also.

